PR #5, opened 2018-06-11
docs(instantsearch.js): update link to loading

bobylito:


Haroenv:
seems like the tests don't pass regardless of this PR @JonasBa

bobylito:
Should we merge that one?

PR #11, opened 2018-09-11
docs(examples): update e-commerce dataset

mthuret:


bobylito:
@mthuret The index is provided through an environment variable. I'm wondering where it is set and if we need to have the acknowledgement line.

mthuret:
maybe @JonasBa can help on this one?

JonasBa:
At the time when I was testing it I used a few different indexe, but this should be hardcode as it's not necessary anymore.

PR #12, opened 2018-09-13
Add onload listener for the script

eele94:
I had some problems with async loading of the file so I added an onload event.

An other fix would be to include the script in my local dependencies. Only no NPM package for the source atm?

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,13 +23,15 @@ search analytics.
 
 ```html
   <script>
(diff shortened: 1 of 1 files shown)


Haroenv:
@JonasBa, you know more about that decision, it should have a queue for calls that are done before the library is initialised? 

It's published on npm though: `search-insights@0.0.15`

JonasBa:
@eele94 so sorry I didn't see this. I'm curious what's the use case here? Why do you need the onload? Indeed as @Haroenv pointed out, there is a queue that's processed before the load, so if you use the library there it might interfere. I'm not a fan of onload here because it also pollutes the global namespace.

tkrugg:
closing this as no response.
@eele94 feel free to reopen if you want to further discuss.

PR #14, opened 2018-11-12
fix(tests): properly start/stop testServer

tkrugg:
I noticed the server was started almost manually before calling jest et not properly stopped after tests have finished.
I think a proper way would be to start and stop the server in beforeAll/afterAll hooks.

This is what this PR is doing.
We can write more tests now and now mode EADDRINUSE when running tests twice

### DIFF
--- a/lib/__tests__/harvest.test.ts
+++ b/lib/__tests__/harvest.test.ts
@@ -1,3 +1,4 @@
+import * as testServer from '../../server/server.js';
 import AlgoliaInsights from './../insights'
 const puppeteer = require('puppeteer');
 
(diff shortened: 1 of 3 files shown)


lib/__tests__/harvest.test.ts
```diff
@@ -102,7 +106,7 @@ describe('Integration tests', () => {
     expect(data).toHaveProperty('queryID')
     expect(requestClick.queryID).toBe(data.queryID)
 
-    expect(requestClick.objectID).toBe('81')
+    expect(requestClick).toHaveProperty('objectID')
```
tkrugg: I've also changed this tests because I thought it made this tests to tightly coupled to one particular index.

Haroenv:
```
/home/travis/build/algolia/search-insights.js/node_modules/html-replace-webpack-plugin/index.js:10
      throw new Error('Both `pattern` and `replacement` options must be defined!')
```

tkrugg:
closing this one in favour of #15

PR #13, opened 2018-11-12
fix(test): remove double import

tkrugg:

PR #16, opened 2018-11-14
fix(test): remove double import

tkrugg:


tkrugg:
was approved here https://github.com/algolia/search-insights.js/pull/13

PR #20, opened 2018-11-28
chore(tools): tslint and prettier

tkrugg:
depends on: https://github.com/algolia/search-insights.js/pull/19

- install prettier, tsling and related config packages 
- removed eslint and related plugin packages
- install tslint and added tslint config (followed the tsconfig from [angular-instantsearch](https://github.com/algolia/angular-instantsearch/blob/develop/tslint.json))

I think it's best to introduce them now rather than later, it helps me go faster.


dist/search-insights.min.js
```diff
@@ -1 +1 @@
...
...t=e.queryID;if("function"!=typeof this.getQueryID||t||(t=this.getQueryID()||this._lastQueryID),!t)throw new Error("No queryID was retrieved, please check the implementation and provide either a getQueryID function\n    or call the search method that will return the queryID parameter");this.storageManager.storeClick(e.objectID,t);var n=Object.assign({},e,{queryID:t});this.sendEvent("click",n)}.bind(this),this.conversion=function(e){if(!this._hasCredentials)throw new Error("Before calling any methods on the analytics, you first need to call the 'init' function with applicationID and apiKey parameters");if(!e)throw new Error("No parameters were sent to conversion event, please provide an objectID");if(!e.objectID)throw new Error("No objectID was sent to conversion event, please provide an objectID");var t=e.queryID?e.queryID:this.storageManager.getConversionObjectID(e.objectID),n=Object.assign(e,{queryID:t});if(!t)throw new Error("No queryID was retrieved, please check the implementation and provide either a getQueryID function\n    or call the conversion method that will return the queryID parameter");this.sendEvent("conversion",n)}.bind(this),this._userID=c(),this.processQueue()}});
```
Haroenv: I wonder why the dist is tracked on GitHub 🤔

lib/__tests__/conversion.test.ts
```diff
@@ -1,57 +1,57 @@
-import AlgoliaInsights from '../insights';
+import AlgoliaInsights from "../insights";
 
 const credentials = {
-  apiKey: 'test',
-  applicationID: 'test'
```
Haroenv: I don't really care much, but it seems like the diff would be smaller if you use `singleQuote: true` in the prettier config

tkrugg: you're right, let's try that

dist/search-insights.min.js
tkrugg: no sure why either. Any insights on this @JonasBa ?

JonasBa: Yeah, there's no need. We can remove it.

lib/__tests__/conversion.test.ts
tkrugg: I tried the singleQuote option of prettier, and I faced issues:
It does transform most of the strings to singleQuote, unless it results in to many escapes, in case which is falls back again to double quotes (Behaviour is explain [here](https://github.com/prettier/prettier/issues/2338#issuecomment-311776517)) 

In shorts: singleQuote:true leads prettier mixe quotes 😨 and tslint is not happy

dist/search-insights.min.js
tkrugg: will do it in my next PR, stay tuned :)

tkrugg: done here: https://github.com/algolia/search-insights.js/pull/21

PR #21, opened 2018-11-30
fix(build): remove dist/ from codebase

tkrugg:
- remove dist/ from codebase
- added dist/ to .gitignore
- added dist to artifacts paths in .travis (so that the build bundle is still stored somewhere)


Haroenv:
Same comment as @samouss :)

tkrugg:
@samouss @Haroenv 
Before this PR, the bundle was versioned (and therefore stored somehow)
Since i'm removing the bundle from the version system, I wanted to keep it stored somehow.

Not sure it's necessary. Let me know your thoughts

PR #22, opened 2018-11-30
feat(v2): implement search params in sendEvent

tkrugg:
This PR adds to sendEvent everything it needs to make search related tracking (click and conversion)

### DIFF
--- a/examples/async/async.js
+++ b/examples/async/async.js
@@ -168,8 +168,8 @@ document.addEventListener('click', (e) => {
     aa('click', {
       eventName: "hit-clicked",
       indexName: process.env.INDEX_NAME,
-      objectID: e.target.getAttribute('data-objectid'),
-      position: parseInt(e.target.getAttribute('data-position'))
+      objectID: [e.target.getAttribute('data-objectid')],
+      position: [parseInt(e.target.getAttribute('data-position'))]
     })
 
   } else if(e.target.matches('.button-convert')) {
@@ -177,7 +177,7 @@ document.addEventListener('click', (e) => {
     aa('conversion', {
       eventName: "hit-converted",
       indexName: process.env.INDEX_NAME,
-      objectID: e.target.getAttribute('data-objectid')
+      objectID: [e.target.getAttribute('data-objectid')]
     })
   }
 })
--- a/examples/async/async.js
+++ b/examples/async/async.js
@@ -2,7 +2,7 @@ require('./async.scss');
 
 import instantsearch from 'instantsearch.js';
 
-import { 
+import {
   searchBox,
   stats,
   hits,
@@ -166,6 +166,8 @@ document.addEventListener('click', (e) => {
   if(e.target.matches('.button-click')) {
 
     aa('click', {
+      eventName: "hit-clicked",
+      indexName: process.env.INDEX_NAME,
       objectID: e.target.getAttribute('data-objectid'),
       position: parseInt(e.target.getAttribute('data-position'))
     })
@@ -173,22 +175,8 @@ document.addEventListener('click', (e) => {
   } else if(e.target.matches('.button-convert')) {
 
     aa('conversion', {
-      objectID: e.target.getAttribute('data-objectid')
-    })
-  }
-})
-
-document.addEventListener('contextmenu', (e) => {
-  if(e.target.matches('.button-click')) {
-
-    window.aa('click', {
-      objectID: e.target.getAttribute('data-objectid'),
-      position: e.target.getAttribute('data-position')
-    })
-
-  } else if(e.target.matches('.button-convert')) {
-
-    window.aa('conversion', {
+      eventName: "hit-converted",
+      indexName: process.env.INDEX_NAME,
       objectID: e.target.getAttribute('data-objectid')
     })
   }
--- a/examples/instantsearch/instantsearchExample.js
+++ b/examples/instantsearch/instantsearchExample.js
@@ -2,7 +2,7 @@ require('./instantsearchExample.scss');
 
 import instantsearch from 'instantsearch.js';
 
-import { 
+import {
   searchBox,
   stats,
   hits,
@@ -164,6 +164,8 @@ document.addEventListener('click', (e) => {
   if(e.target.matches('.button-click')) {
 
     window.aa('click', {
+      eventName: "hit-clicked",
+      indexName: process.env.INDEX_NAME,
       objectID: e.target.getAttribute('data-objectid'),
       position: parseInt(e.target.getAttribute('data-position'))
     })
@@ -171,22 +173,8 @@ document.addEventListener('click', (e) => {
   } else if(e.target.matches('.button-convert')) {
 
     window.aa('conversion', {
-      objectID: e.target.getAttribute('data-objectid')
-    })
-  }
-})
-
-document.addEventListener('contextmenu', (e) => {
-  if(e.target.matches('.button-click')) {
-
-    window.aa('click', {
-      objectID: e.target.getAttribute('data-objectid'),
-      position: e.target.getAttribute('data-position')
-    })
-
-  } else if(e.target.matches('.button-convert')) {
-
-    window.aa('conversion', {
+      eventName: "hit-converted",
+      indexName: process.env.INDEX_NAME,
       objectID: e.target.getAttribute('data-objectid')
     })
   }
--- a/examples/instantsearch/instantsearchExample.js
+++ b/examples/instantsearch/instantsearchExample.js
@@ -166,8 +166,8 @@ document.addEventListener('click', (e) => {
     window.aa('click', {
       eventName: "hit-clicked",
       indexName: process.env.INDEX_NAME,
-      objectID: e.target.getAttribute('data-objectid'),
-      position: parseInt(e.target.getAttribute('data-position'))
+      objectID: [e.target.getAttribute('data-objectid')],
+      position: [parseInt(e.target.getAttribute('data-position'))]
     })
 
   } else if(e.target.matches('.button-convert')) {
@@ -175,7 +175,7 @@ document.addEventListener('click', (e) => {
     window.aa('conversion', {
       eventName: "hit-converted",
       indexName: process.env.INDEX_NAME,
-      objectID: e.target.getAttribute('data-objectid')
+      objectID: [e.target.getAttribute('data-objectid')]
     })
   }
 })
--- a/examples/lunr/lunr.js
+++ b/examples/lunr/lunr.js
@@ -77,6 +77,8 @@ document.addEventListener('click', (e) => {
   if(e.target.matches('.button-click')) {
 
     window.aa('click',{
+      eventName: "hit-clicked",
+      indexName: process.env.INDEX_NAME,
       objectID: e.target.getAttribute('objectid'),
       position: e.target.getAttribute('position')
     })
@@ -84,6 +86,8 @@ document.addEventListener('click', (e) => {
   } else if(e.target.matches('.button-convert')) {
 
     window.aa('conversion',{
+      eventName: "hit-converted",
+      indexName: process.env.INDEX_NAME,
       objectID: e.target.getAttribute('objectid')
     })
   }
--- a/examples/lunr/lunr.js
+++ b/examples/lunr/lunr.js
@@ -79,8 +79,8 @@ document.addEventListener('click', (e) => {
     window.aa('click',{
       eventName: "hit-clicked",
       indexName: process.env.INDEX_NAME,
-      objectID: e.target.getAttribute('objectid'),
-      position: e.target.getAttribute('position')
+      objectID: [e.target.getAttribute('objectid')],
+      position: [e.target.getAttribute('position')]
     })
 
   } else if(e.target.matches('.button-convert')) {
@@ -88,7 +88,7 @@ document.addEventListener('click', (e) => {
     window.aa('conversion',{
       eventName: "hit-converted",
       indexName: process.env.INDEX_NAME,
-      objectID: e.target.getAttribute('objectid')
+      objectID: [e.target.getAttribute('objectid')]
     })
   }
 })
--- a/examples/react-instantsearch/src/App.js
+++ b/examples/react-instantsearch/src/App.js
@@ -29,6 +29,8 @@ const Hits = connectHits(
           <button
             onClick={() => {
               window.aa('click', {
+                eventName: "hit-clicked",
+                indexName: process.env.INDEX_NAME,
                 objectID: hit.objectID,
                 position:
                   searchResults.hitsPerPage * searchResults.page + index + 1,
@@ -40,6 +42,8 @@ const Hits = connectHits(
           <button
             onClick={() => {
               window.aa('conversion', {
+                eventName: "hit-converted",
+                indexName: process.env.INDEX_NAME,
                 objectID: hit.objectID,
               });
             }}
--- a/examples/react-instantsearch/src/App.js
+++ b/examples/react-instantsearch/src/App.js
@@ -31,9 +31,9 @@ const Hits = connectHits(
               window.aa('click', {
                 eventName: "hit-clicked",
                 indexName: process.env.INDEX_NAME,
-                objectID: hit.objectID,
+                objectID: [hit.objectID],
                 position:
-                  searchResults.hitsPerPage * searchResults.page + index + 1,
+                  [searchResults.hitsPerPage * searchResults.page + index + 1],
               });
             }}
           >
@@ -44,7 +44,7 @@ const Hits = connectHits(
               window.aa('conversion', {
                 eventName: "hit-converted",
                 indexName: process.env.INDEX_NAME,
-                objectID: hit.objectID,
+                objectID: [hit.objectID],
               });
             }}
           >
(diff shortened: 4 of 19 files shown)


examples/async/async.js
```diff
@@ -166,29 +166,17 @@ document.addEventListener('click', (e) => {
...
 
   } else if(e.target.matches('.button-convert')) {
 
     aa('conversion', {
-      objectID: e.target.getAttribute('data-objectid')
-    })
-  }
-})
-
-document.addEventListener('contextmenu', (e) => {
```
tkrugg: I simplified the example and removed the context-menu listener here.

examples/instantsearch/instantsearchExample.js
```diff
@@ -164,29 +164,17 @@ document.addEventListener('click', (e) => {
...
 
   } else if(e.target.matches('.button-convert')) {
 
     window.aa('conversion', {
-      objectID: e.target.getAttribute('data-objectid')
-    })
-  }
-})
-
-document.addEventListener('contextmenu', (e) => {
```
tkrugg: I simplified the example and removed the context-menu listener here.

lib/utils/index.ts
```diff
@@ -0,0 +1,5 @@
+// use theses type checking helpers to avoid mistyping "undefind", I mean "undfined"
+export const isUndefined = value => typeof value === "undefined";
+export const isString = value => typeof value === "string";
+export const isNumber = value => typeof value === "number";
+export const isFunction = value => typeof value === "function";
```
tkrugg: I could have used lodash for this but it seemed like too much code: (eg. https://github.com/lodash/lodash/blob/master/isFunction.js)

examples/async/async.js
```diff
@@ -166,29 +166,17 @@ document.addEventListener('click', (e) => {
   if(e.target.matches('.button-click')) {
 
     aa('click', {
+      eventName: "hit-clicked",
+      indexName: process.env.INDEX_NAME,
       objectID: e.target.getAttribute('data-objectid'),
       position: parseInt(e.target.getAttribute('data-position'))
```
samouss: Careful inside the example because I don't think the provided position is the correct one. The one we get here is the relative but IIRC the one we have to provide is the absolute, correct?

examples/react-instantsearch/src/App.js
```diff
@@ -29,6 +29,8 @@ const Hits = connectHits(
           <button
             onClick={() => {
               window.aa('click', {
+                eventName: "hit-clicked",
+                indexName: process.env.INDEX_NAME,
```
samouss: We should avoid to document this usage because the widget can be scoped inside an `Index` component that change the index currently targeted. We should document the "best" way to access all those information.

examples/vue-instantsearch/src/App.vue
```diff
@@ -97,11 +97,20 @@ export default {
       // prefix = page (0-based) * hitsPerPage
       // position is 1-based
       const position = this.prefix + index + 1;
-      window.aa("click", { objectID, position });
+      window.aa("click", {
+        eventName: "hit-clicked",
+        indexName: process.env.INDEX_NAME,
```
samouss: We should document the "best" way to access all those information.

examples/lunr/lunr.js
```diff
@@ -77,13 +77,17 @@ document.addEventListener('click', (e) => {
   if(e.target.matches('.button-click')) {
 
     window.aa('click',{
+      eventName: "hit-clicked",
+      indexName: process.env.INDEX_NAME,
```
samouss: We should document the "best" way to access all those information.

lib/__tests__/_sendEvent.test.ts
```diff
@@ -41,17 +43,21 @@ describe("sendEvent", () => {
...
       });
       expect(XMLHttpRequest.send).toHaveBeenCalledTimes(1);
       const payload = JSON.parse(XMLHttpRequest.send.mock.calls[0][0]);
       expect(payload).toEqual({
         events: [
           expect.objectContaining({
             eventType: "click",
-            objectID: "1",
-            userID: "42",
-            timestamp: expect.any(Number)
```
samouss: timestamp is not provided anymore?

lib/__tests__/_sendEvent.test.ts
```diff
@@ -41,17 +43,21 @@ describe("sendEvent", () => {
...
+            eventName: "my-event",
+            indexName: "my-index",
+            objectID: ["1"],
```
samouss: Since the value is an array shouldn't be `objectIDs`?

lib/__tests__/_sendEvent.test.ts
```diff
@@ -41,17 +43,21 @@ describe("sendEvent", () => {
...
+            position: [3]
```
samouss: Since the value is an array shouldn't be `positions`?

lib/__tests__/_sendEvent.test.ts
```diff
@@ -84,20 +94,189 @@ describe("sendEvent", () => {
...
         events: [
           expect.objectContaining({
             eventType: "click",
-            objectID: "1",
-            userID: "42",
+            eventName: "my-event",
+            indexName: "my-index",
+            objectID: ["1"],
+            position: [3],
+            userID: "mock-user-id",
```
samouss: Why do we have the `userID` for the beacon but not for the XMLHttpRequest? Rather than using `objectContaining` I would use something more strict like `toEqual` since we assert the full payload.

lib/__tests__/_sendEvent.test.ts
```diff
@@ -84,20 +94,189 @@ describe("sendEvent", () => {
...
+    it("should throw if no eventName passed", () => {
+      expect(() => {
+        (AlgoliaInsights as any).sendEvent("click", {
+          indexName: "my-index"
+        });
+      }).toThrowErrorMatchingInlineSnapshot(
+        `"expected required parameter \`eventName\` to be a string"`
+      );
+    });
+    it("should throw if no eventName passed", () => {
```
samouss: We could use a different title for the the assertion.

lib/__tests__/_sendEvent.test.ts
```diff
@@ -84,20 +94,189 @@ describe("sendEvent", () => {
...
+    it("should throw if no indexName passed", () => {
+      expect(() => {
+        (AlgoliaInsights as any).sendEvent("click", {
+          eventName: "my-event"
+        });
+      }).toThrowErrorMatchingInlineSnapshot(
+        `"expected required parameter \`indexName\` to be a string"`
+      );
+    });
+    it("should throw if no indexName is not a string", () => {
```
samouss: We could use a different title for the the assertion.

lib/__tests__/_sendEvent.test.ts
```diff
@@ -84,20 +94,189 @@ describe("sendEvent", () => {
...
+            position: [3]
+          })
+        ]
+      });
+    });
+    it("should support multiple objectID and position", () => {
+      (AlgoliaInsights as any).sendEvent("click", {
+        eventName: "my-event",
+        indexName: "my-index",
+        objectID: ["1", "2"],
```
samouss: What about support only the plural form?

lib/_sendEvent.ts
```diff
@@ -1,37 +1,79 @@
...
-export type InsightsEventTypes = 'click' | 'conversion';
+export type InsightsEventType = "click" | "conversion";
 export type InsightsEvent = {
-  eventType: InsightsEventTypes;
+  eventType: InsightsEventType;
+
   eventName: string;
   userID: string;
-  timestamp: number;
+  timestamp: number; // TODO: we should allow Date object as well?
```
samouss: Don't think so, we should stick with a number IMO.

lib/_sendEvent.ts
```diff
@@ -1,37 +1,79 @@
...
   indexName: string;
+
+  queryID?: string;
+  objectID?: (string | number) | (string | number)[];
```
samouss: Why do we want to support both form?

lib/click.ts
```diff
@@ -1,8 +1,12 @@
...
+  eventName: string;
+  userID: string;
+  timestamp: number;
+  indexName: string;
 
-export interface InsightsSearchClickEvent extends InsightsEvent {
   objectID: string | number;
   position: number;
+
   queryID?: string;
```
samouss: From what I understand inside the code the `queryID` is required.

lib/click.ts
```diff
@@ -15,15 +19,18 @@ export function click(params: InsightsSearchClickEvent) {
     throw new Error(
       "Before calling any methods on the analytics, you first need to call the 'init' function with applicationID and apiKey parameters"
     );
-  } else if (!params) {
+  }
+  if (!params) {
```
samouss: The error message is outdated, more parameters are now required.

lib/_sendEvent.ts
```diff
@@ -1,37 +1,79 @@
...
+import { isNumber, isUndefined, isString, isFunction } from "./utils/index";
 
 declare var process: {
   env: {
     NODE_ENV: string;
   };
 };
 
-export type InsightsEventTypes = 'click' | 'conversion';
+export type InsightsEventType = "click" | "conversion";
```
samouss: Why we dropped the plural form?

lib/utils/index.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+describe("isUndefined", () => {
+  const tests = [
+    { input: 42, expected: false },
+    { input: "a string", expected: false },
+    { input: undefined, expected: true },
+    { input: () => null, expected: false }
+  ];
+  for (const { input, expected } of tests) {
+    it(`should return ${expected} when passed ${input}`, () => {
```
samouss: You can use a Jest [`each` ](https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout).

lib/conversion.ts
```diff
@@ -1,6 +1,9 @@
-import { InsightsEvent } from "./_sendEvent";
+export interface InsightsSearchConversionEvent {
+  eventName: string;
+  userID: string;
+  timestamp: number;
+  indexName: string;
 
-export interface InsightsSearchConversionEvent extends InsightsEvent {
   objectID: string | number;
```
samouss: We cannot provide multiple `objectID`?

lib/click.ts
```diff
@@ -1,8 +1,12 @@
-import { InsightsEvent } from "./_sendEvent";
+export interface InsightsSearchClickEvent {
+  eventName: string;
+  userID: string;
+  timestamp: number;
+  indexName: string;
 
-export interface InsightsSearchClickEvent extends InsightsEvent {
   objectID: string | number;
   position: number;
```
samouss: We cannot provide multiple `position` & `objectID`?

lib/_sendEvent.ts
tkrugg: @Haroenv suggested to change it here: https://github.com/algolia/search-insights.js/pull/19#discussion_r237392498

What's the best practice for naming enum types?

lib/click.ts
tkrugg: I thought I would do that later but I can add it here

lib/conversion.ts
tkrugg: I thought I would do that later but I can add it here to this PR

examples/async/async.js
JonasBa: That's cool, do you even think it's worth having a context-menu example? I think it makes sense, especially as some people I imagine search for something and just open everything in a new tab. In that case we might be missing data and it could be quite significant. I would advocate to have this in our doc tho, wdyt?

lib/_sendEvent.ts
samouss: Good question, I quickly search but didn't find anything.

examples/async/async.js
tkrugg: > I think it makes sense, especially as some people I imagine search for something and just open everything in a new tab

oh! you're right. Now I understand the point of having that listener.
We should definitely add it to our docs!

I wonder if there's anything more precise: receiving a context menu event might not mean the user open it in a new tab. I guess it was hard to find a better solution?

lib/__tests__/_sendEvent.test.ts
tkrugg: it is but I'm purposely not checking that here.
I wanted to keep this about the payload passed to `XMLHttpRequest.send` so I added only required parameters.
This is why i've removed userID and timestamp

To be fully consistent, I'll also remove objectID and positions (as they are also optional)

lib/__tests__/_sendEvent.test.ts
tkrugg: The endpoint accepts only `objectID: string[]`.
In the RFC, we agreed that we'll expose this as `objectIDs: string[]` because it gives a better hint on the nature of it.
```js
AlgoliaInsights.click({  objectIDs: ["1"], positions: [ 1 ] }) 
// which will call 
AlgoliaInsights._sendEvent({ eventType: "click", objectID: ["1"], position: [ 1 ] }) // signature is exactly matching payload requirement on the endpoint
// which will call 
XMLHttpRequest.send({ eventType: "click", objectID: ["1"], position: [ 1 ] })
```
If this is clear, please share your thoughts 🙂

lib/__tests__/_sendEvent.test.ts
tkrugg: same answer as objectID/objectIDs: https://github.com/algolia/search-insights.js/pull/22/files/b2e374f96fda0c2b903194f879974b41f62618b9#r238710452

lib/__tests__/_sendEvent.test.ts
tkrugg: since the content of the payload is checked in details in other tests, i'll simplify this one and only keep required param
```js
{
    eventType: "click",
    eventName: "my-event",
    indexName: "my-index",	            
}
```

lib/_sendEvent.ts
tkrugg: Actually not, let's remove this :)

lib/click.ts
tkrugg: will fix that

examples/react-instantsearch/src/App.js
tkrugg: I'll have to get rid of process.env.INDEX_NAME in the close future
The ultimate best way to use insights client will come with the instantsearch integration.
I filed an issue for that

examples/vue-instantsearch/src/App.vue
tkrugg: issue filed [here](https://github.com/algolia/search-insights.js/issues/25)

examples/lunr/lunr.js
tkrugg: issue filed [here](https://github.com/algolia/search-insights.js/issues/25)

lib/__tests__/_sendEvent.test.ts
tkrugg: you mean wrap them in a separate "describe" block?

lib/__tests__/_sendEvent.test.ts
tkrugg: this is answered [here](https://github.com/algolia/search-insights.js/pull/22#discussion_r238710452)

examples/async/async.js
samouss: Don't use the built-in widgets but only the connectors.

lib/_sendEvent.ts
tkrugg: We can keep it singular like every other type

examples/async/async.js
tkrugg: ok, i've opened an issue so that we don't forget it: https://github.com/algolia/search-insights.js/issues/26

lib/__tests__/_sendEvent.test.ts
tkrugg: It's done. It should be more minimal now.

examples/async/async.js
tkrugg: opened a PR on top for that https://github.com/algolia/search-insights.js/pull/31

lib/__tests__/_sendEvent.test.ts
samouss: Both of the assertions inside the `describe` have the exact same name. It's easier to follow if they have different name. You don't test the exact same thing in both assertions.

lib/__tests__/_sendEvent.test.ts
```diff
@@ -68,36 +67,192 @@ describe("sendEvent", () => {
...
+          position: [3]
+        });
+      }).toThrowErrorMatchingInlineSnapshot(
+        `"Cannot use \`position\` without providing \`objectID\`"`
+      );
+    });
+  });
+
+  describe("timestamp", () => {
+    it("should add a timestamp if not provided", () => {
```
samouss: We dropped the support of this feature for v1. Is it the same for v2?

examples/async/async.js
tkrugg: Note that i'm not using connectors but transformData as discussed with Marie

lib/__tests__/_sendEvent.test.ts
tkrugg: my bad, 😅  fixed here: https://github.com/algolia/search-insights.js/pull/33

PR #24, opened 2018-12-04
fix(sendEvent): remove enforced Date.now() timestamp

tkrugg:
⚠️  This will affect all endpoints: `/1/click`, `/1/conversion`

- [x] confirm with @sfriquet timestamp is not needed for theses endpoints?

This will be merged inside master and released in 0.0.16.

Motivation:
1. This is not needed as it turns out the server provides it by default.
2. `Date.now()` has the value of the end users's computer clock. It may or it may not be on time. If it's in the future, the server will reject it with a 422.

Drawbacks: None


lib/_sendEvent.ts
```diff
@@ -10,7 +10,7 @@ declare var process : {
 
 export type AnalyticsEvent = {
   userID: string;
-  timestamp: number;
+  timestamp?: number;
```
samouss: Why do we want to keep the `timestamp` optional since we don't use it anymore?

tkrugg: My intention was that timestamp can still provided, only now we don't override it with `Date.now()`.

sfriquet:
As long as you send the events right away this is fine.

Also note that /1/click and /2/conversion are deprecated.

PR #31, opened 2018-12-07
fix(examples): use aboslute position in examples

tkrugg:
Events of the API expect an absolute position.
I changed the examples to conform to it.

follow up on this comment from @samouss


examples/async/async.js
```diff
@@ -80,6 +83,12 @@ search.addWidget(
       item: hitTemplate
     },
     transformData: function(hit) {
+      var result = search.helper.lastResults;
+      var offset = result.hitsPerPage * result.page;
+
+      hit.queryID = result.queryID;
+      hit.hitPosition = offset + hit.__hitIndex + 1;
```
Haroenv: these two should probably have two underscores so that they don't collapse with user's attributes

examples/async/async.js
```diff
@@ -170,13 +179,15 @@ document.addEventListener("click", e => {
     aa("click", {
       eventName: "hit-clicked",
       index: process.env.INDEX_NAME,
+      queryID: e.target.getAttribute("data-queryid"),
```
samouss: `data-query-id`

examples/async/async.js
```diff
@@ -170,13 +179,15 @@ document.addEventListener("click", e => {
...
       objectIDs: [e.target.getAttribute("data-objectid")],
```
samouss: `data-object-id`

examples/async/async.js
tkrugg: done

examples/async/async.js
tkrugg: done

examples/lunr/lunr.js
```diff
@@ -87,14 +87,14 @@ document.addEventListener("click", e => {
     window.aa("click", {
       eventName: "hit-clicked",
       index: process.env.INDEX_NAME,
-      objectID: [e.target.getAttribute("objectid")],
-      position: [e.target.getAttribute("position")]
+      objectIDs: [e.target.getAttribute("objectid")],
```
samouss:
```suggestion
      objectIDs: [e.target.getAttribute("data-object-id")],
```

examples/lunr/lunr.js
```diff
@@ -35,7 +35,7 @@ var hitTemplate = (hit, i) => `
   </div>
   <button objectID="${hit.objectID}" position="${i +
```
samouss: You can change their name to be consistent with the others: `data-xxx`.

examples/lunr/lunr.js
```diff
@@ -87,14 +87,14 @@ document.addEventListener("click", e => {
...
-      position: [e.target.getAttribute("position")]
+      objectIDs: [e.target.getAttribute("objectid")],
+      positions: [e.target.getAttribute("position")]
     });
   } else if (e.target.matches(".button-convert")) {
     window.aa("conversion", {
       eventName: "hit-converted",
       index: process.env.INDEX_NAME,
-      objectID: [e.target.getAttribute("objectid")]
+      objectIDs: [e.target.getAttribute("objectid")]
```
samouss:
```suggestion
      objectIDs: [e.target.getAttribute("data-object-id")]
```

examples/lunr/lunr.js
tkrugg: Thanks, missed that one!
i've added a commit to tackle it
also fixed `position -> data-position`

PR #34, opened 2018-12-07
feat(filters): add lower level support for filters

tkrugg:


lib/_sendEvent.ts
```diff
@@ -73,14 +79,30 @@ export function sendEvent(
...
+    }
+    if (!Array.isArray(eventData.filters)) {
+      throw TypeError("expected optional parameter `filters` to be an array");
+    }
+    event.filters = eventData.filters;
+  }
+
+  if (isUndefined(eventData.objectIDs) && isUndefined(eventData.filters)) {
+    throw new Error("expected either `objectIDs` or `filters` to be provided");
+  }
```
samouss: The function became a bit dense. In the future if more parameter are added it will become hard to follow IMO. A possible solution would be to externalise all the validation functions and perform them as soon as possible. It has the advantage to be focused on each method. The downside is that we have to repeat the same validation for the common parameters but the tradeoff seems acceptable to me. WDYT @tkrugg? We can do it later though no problem at the moment.

tkrugg: I agree. 
I've opened an issue for it here: https://github.com/algolia/search-insights.js/issues/44

PR #37, opened 2018-12-10
feat(conversion): add convertedObjectIDs and convertedFilters

tkrugg:
follows up on https://github.com/algolia/search-insights.js/pull/36
Adds generic conversion methods `convertedObjectIDs` and `convertedFilters`

### DIFF
--- a/lib/__tests__/conversion.test.ts
+++ b/lib/__tests__/conversion.test.ts
@@ -4,13 +4,13 @@ const credentials = {
   apiKey: "test",
   applicationID: "test"
 };
-describe("convertedObjectIDInSearch", () => {
+describe("convertedObjectIDsAfterSearch", () => {
   it("Should throw if no params are sent", () => {
     expect(() => {
       AlgoliaInsights.init(credentials);
-      (AlgoliaInsights as any).convertedObjectIDInSearch();
+      (AlgoliaInsights as any).convertedObjectIDsAfterSearch();
     }).toThrowError(
-      "No params were sent to convertedObjectIDInSearch function, please provide `queryID` and `objectIDs` to be reported"
+      "No params were sent to convertedObjectIDsAfterSearch function, please provide `queryID` and `objectIDs` to be reported"
     );
   });
 
@@ -19,7 +19,7 @@ describe("convertedObjectIDInSearch", () => {
     AlgoliaInsights.init(credentials);
 
     expect(() => {
-      (AlgoliaInsights as any).convertedObjectIDInSearch({
+      (AlgoliaInsights as any).convertedObjectIDsAfterSearch({
         objectIDs: ["12345"]
       });
       expect((AlgoliaInsights as any).sendEvent).not.toHaveBeenCalled();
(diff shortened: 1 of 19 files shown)


lib/__tests__/conversion.test.ts
```diff
@@ -54,3 +54,73 @@ describe("convertedObjectIDInSearch", () => {
...
+      "No params were sent to convertedObjectIDs function, please provide `objectIDs` to be reported"
+    );
+  });
+
+  it("should throw if no objectIDs has been passed", () => {
+    (AlgoliaInsights as any).sendEvent = jest.fn();
+    AlgoliaInsights.init(credentials);
+
+    expect(() => {
+      (AlgoliaInsights as any).convertedObjectIDs({ queryID: "test" });
```
samouss: The `queryID` should not be present if I understood correctly, since we are not inside the context of a search. It should be caught by the TypeScript though but since we cast the object it does not see the error.

lib/__tests__/conversion.test.ts
```diff
@@ -54,3 +54,73 @@ describe("convertedObjectIDInSearch", () => {
...
+      expect((AlgoliaInsights as any).sendEvent).not.toHaveBeenCalled();
+    }).toThrowError(
+      "required objectIDs parameter was not sent, conversion event can not be properly sent without"
+    );
+  });
+
+  it("should send allow passing of queryID", () => {
```
samouss: The description is not related to the test.

lib/__tests__/conversion.test.ts
```diff
@@ -54,3 +54,73 @@ describe("convertedObjectIDInSearch", () => {
...
+  it("should throw if no params are sent", () => {
+    expect(() => {
+      AlgoliaInsights.init(credentials);
+      (AlgoliaInsights as any).convertedFilters();
+    }).toThrowError(
+      "No params were sent to convertedFilters function, please provide `filters` to be reported"
+    );
+  });
+
+  it("should throw if no objectIDs has been passed", () => {
```
samouss: The description is not related to the test.

lib/__tests__/conversion.test.ts
```diff
@@ -54,3 +54,73 @@ describe("convertedObjectIDInSearch", () => {
...
+
+    expect(() => {
+      (AlgoliaInsights as any).convertedFilters({});
+      expect((AlgoliaInsights as any).sendEvent).not.toHaveBeenCalled();
+    }).toThrowError(
+      "required filters parameter was not sent, conversion event can not be properly sent without"
+    );
+  });
+
+  it("should send allow passing of queryID", () => {
```
samouss: The description is not related to the test.

lib/conversion.ts
```diff
@@ -35,3 +35,59 @@ export function convertedObjectIDInSearch(
...
+  if (!params.objectIDs) {
+    throw new Error(
+      "required objectIDs parameter was not sent, conversion event can not be properly sent without"
+    );
+  }
+
+  this.sendEvent("conversion", params as InsightsEvent);
+}
+
+export interface InsightsSearchConversionFiltersEvent {
```
samouss: Why do we have the `Search` inside the name? This event is not tied to a search right?

lib/conversion.ts
```diff
@@ -35,3 +35,59 @@ export function convertedObjectIDInSearch(
 
   this.sendEvent("conversion", params as InsightsEvent);
 }
+
+export interface InsightsSearchConversionObjectIDsEvent {
```
samouss: Why do we have the Search inside the name? This event is not tied to a search right?

lib/__tests__/conversion.test.ts
tkrugg: removed queryID, you're right it has nothing to do there!

lib/conversion.ts
tkrugg: fixed in next-next commit!

PR #38, opened 2018-12-10
feat(view): add viewedObjectIDs and viewedFilters

tkrugg:
This PR adds generic methods `viewedObjectIDs` and `viewedFilters` as specified in the related [RFC](https://github.com/algolia/instantsearch-rfcs/blob/feat/search-insights-client/accepted/search-insights-client.md#detailed-design-by-stories)


tkrugg:
@samouss it's fixed! thanks for the review

PR #40, opened 2018-12-10
feat(init): introduce region param and compute endpointOrigin

tkrugg:
Target story:

> A developer should be able to choose the region he/she wants to use to send events

- we introduce region param
- compute endpointOrigin, but it's still unsued 

will start using it in next PR (there are too tests to migrate so best done in a separate PR)

### DIFF
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -27,4 +27,43 @@ describe("init", () => {
       `"applicationID is missing, please provide it, so we can properly attribute data to your application"`
     );
   });
+  it("should throw if region is other than `de` | `us`", () => {
+    expect(() => {
+      (AlgoliaInsights as any).init({
+        applicationID: "xxx",
+        apiKey: "***",
+        region: "emea"
+      });
+    }).toThrowErrorMatchingInlineSnapshot(
+      `"optional region is incorrect, please provide either one of: de, us."`
+    );
+  });
+  it("should set _applicationID on instance", () => {
+    AlgoliaInsights.init({ apiKey: "***", applicationID: "XXX" });
+    expect(AlgoliaInsights._applicationID).toBe("XXX");
+  });
+  it("should set _apiKey on instance", () => {
+    AlgoliaInsights.init({ apiKey: "***", applicationID: "XXX" });
+    expect(AlgoliaInsights._apiKey).toBe("***");
+  });
+  it("should set _region on instance", () => {
+    AlgoliaInsights.init({ apiKey: "***", applicationID: "XXX", region: "us" });
+    expect(AlgoliaInsights._region).toBe("us");
+  });
+  it("should set _endpointOrigin on instance to https://insights.algolia.io", () => {
+    AlgoliaInsights.init({ apiKey: "***", applicationID: "XXX" });
+    expect(AlgoliaInsights._endpointOrigin).toBe("https://insights.algolia.io");
+  });
+  it("should set _endpointOrigin on instance to https://insights.us.algolia.io if region === 'us'", () => {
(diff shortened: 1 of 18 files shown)


lib/init.ts
```diff
@@ -1,6 +1,13 @@
+import { isUndefined, isString } from "./utils/index";
+
+type InsightRegion = "de" | "us";
+const SUPPORTED_REGIONS = ["de", "us"];
```
samouss: The type is not correct, the compiler won't complain for `['de', 'us', 'whatever']`. You can enforce it with an array of `InsightRegion`.

lib/init.ts
```diff
@@ -12,21 +19,34 @@ export function init(options: InitParams) {
...
+      )}.`
+    );
+  }
 
   this._apiKey = options.apiKey;
   this._applicationID = options.applicationID;
+  this._region = options.region;
+  this._endpointOrigin = options.region
+    ? `https://insights.${options.region}.algolia.io`
+    : "https://insights.algolia.io";
```
samouss: It means that we can't provide a specific endpoint (e.g. testing) unless we override the `_endpointOrigin`. We could use `ENV` variables to define the endpoint rather than having it hard coded in the source code.

lib/init.ts
tkrugg: I don't understand what you mean.
on the playground when I write [this](https://www.typescriptlang.org/play/index.html#src=type%20A%20%3D%20'us'%20%7C%20'de'%3B%0D%0Aconst%20a%3A%20A%20%3D%20'emea'%0D%0A) it complains

lib/init.ts
tkrugg: yes, I didn't want to expose it because it's not part of the spec.
Yes it's a problem for testing, but there is a workaround so I didn't mind. 

If we expose something to set an endpoint, I suspect we'd have to document how it overrides regions. I thought it was simpler this way.

samouss: I was not talking about exposing an API but more a replacement at build step something like that.

lib/__tests__/init.test.ts
```diff
@@ -27,4 +27,43 @@ describe("init", () => {
...
+        apiKey: "***",
+        region: "emea"
+      });
+    }).toThrowErrorMatchingInlineSnapshot(
+      `"optional region is incorrect, please provide either one of: de, us."`
+    );
+  });
+  it("should set _applicationID on instance", () => {
+    AlgoliaInsights.init({ apiKey: "***", applicationID: "XXX" });
+    expect(AlgoliaInsights._applicationID).toBe("XXX");
```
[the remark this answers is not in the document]
tkrugg: There are integration tests checking this. So we can already delete theses tests if we think they are not useful.
However setting `_applicationID` (and other internal data)  is the important behaviour of init. Other functions rely on it [sendEvent](https://github.com/algolia/search-insights.js/blob/develop/lib/_sendEvent.ts#L84). So this unit tests is here to ensure this behaviour is preserved and is import imho.

lib/init.ts
samouss: You have to add `const SUPPORTED_REGIONS: InsightRegion[]  = ["de", "us"];`

lib/init.ts
```diff
@@ -1,6 +1,13 @@
+import { isUndefined, isString } from "./utils/index";
+
+type InsightRegion = "de" | "us";
+const SUPPORTED_REGIONS: InsightRegion[] = ["de", "us"];
+
 export interface InitParams {
   apiKey: string;
   applicationID: string;
+  region?: InsightRegion;
+  endpointOrigin?: string;
```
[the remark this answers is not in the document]
tkrugg: i'll fix that in a next PR

PR #41, opened 2018-12-10
feat(region): use region based endpoint

### LINKED ISSUE #23: use replace plugin instead of conditions on process.env
This follows up on this comment:
https://github.com/algolia/search-insights.js/pull/19#discussion_r237895787

tkrugg:
Target story:

> A developer should be able to choose the region he/she wants to use to send events

In the last PR https://github.com/algolia/search-insights.js/pull/40 we introduce region param and compute the endpoint origin.
In this PR, we use the endpoint.

i've also removed the process.env.NODE_ENV check from the codebase (partly addressing #23)

### DIFF
--- a/examples/async/async.js
+++ b/examples/async/async.js
@@ -176,7 +176,7 @@ search.start();
 
 document.addEventListener("click", e => {
   if (e.target.matches(".button-click")) {
-    aa("clickedObjectIDInSearch", {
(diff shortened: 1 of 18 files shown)


tkrugg:
Yes you're right! i've added a commit to address that.
There are still some process.env.NODE_ENV in the examples (examples are a separate thing, they are compiled by webpack, i'll close #23 when they're done)

PR #42, opened 2018-12-10
fix(userToken): migrate userID => userToken

tkrugg:
endpoint now is expects userToken instead of userID.
This is a simple search and replace over all the code base

### DIFF
--- a/lib/_cookieUtils.ts
+++ b/lib/_cookieUtils.ts
@@ -64,7 +64,6 @@ const createUUID = () => {
   });
 };
 
-const sessionID = createUUID;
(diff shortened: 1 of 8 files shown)


lib/_cookieUtils.ts
```diff
@@ -65,6 +65,6 @@ const createUUID = () => {
 };
 
 const sessionID = createUUID;
```
samouss: `sessionToken`?

tkrugg: actually i'll remove it altogether because it's unused and I don't think we need to expose it externally.

PR #48, opened 2018-12-11
feat(userHasOptOut): introduce userHasOptOutBoolean optional param

tkrugg:
Target story:
> A developer should be able to opt-out easily from the usage of our insights library library will be explicitly loaded by the user

In this PR:
- add userHasOptOut (default: false)
- block event reporting if userHasOptOutBoolean: true


Haroenv:
correct spelling is `userHasOptedOut`

PR #57, opened 2018-12-14
fix(types): remove Search from types unrelated to search.

tkrugg:
Follows up on this feedback from @samouss in https://github.com/algolia/search-insights.js/pull/37#pullrequestreview-184589476 and https://github.com/algolia/search-insights.js/pull/38#pullrequestreview-184592263.

- renamed a few tests as they either contain a typo or unrelated to their assertions
- removed Search from types that are unrelated to search, namely: `InsightsConversionFiltersEvent`, `InsightsConversionObjectIDsEvent`, `InsightsViewFiltersEvent ` and `InsightsViewObjectIDsEvent`.


lib/insights.ts
```diff
@@ -75,21 +75,21 @@ class AlgoliaAnalytics {
...
   public init: (params: InitParams) => void;
   public initSearch: (params: InitSearchParams) => void;
-  public clickedObjectIDsAfterSearch: (params?: InsightsSearchClickEvent) => void;
+  public clickedObjectIDsAfterSearch: (
+    params?: InsightsSearchClickEvent
+  ) => void;
   public clickedObjectIDs: (params?: InsightsClickObjectIDsEvent) => void;
   public clickedFilters: (params?: InsightsClickFiltersEvent) => void;
   public convertedObjectIDsAfterSearch: (
     params?: InsightsSearchConversionEvent
```
samouss: Is there a reason why the names `InsightsSearchConversionEvent` and `InsightsConversionObjectIDsEvent` are not consistent (not the `ObjectIDs` part)? It seems that for the others events the name is consistent.

tkrugg: would you expect something like `InsightsConversionSearchObjectIDsEvent`?

Search always use objectIDs, it can be considered as implied by the `Search` part

samouss: One or the other (doesn't matter) but we should be consistent.

PR #58, opened 2018-12-14
chore(npmignore): lighten package

tkrugg:
The package currently contains a lot of files.
This PR aims to keep only the following.

- [ ] tslint.json
- [ ] .nvmrc
- [ ] .travis.yml
- [x] LICENSE.md
- [x] README.md
- [ ] config
- [ ] dev.sh
- [x] dist
- [ ] examples
- [x] lib
- [x] package.json
- [ ] pre-deploy.sh
- [ ] rollup.config.js
- [ ] server
- [x] tsconfig.json
- [x] yarn.lock

In my opinion even yarn.lock is unecessary.
I don't know yet about tsconfig.json


tkrugg:
@samouss ok will do. 
Also do you agree with the list of files checked? Should tsconfig.json be included?

samouss:
Nope it should not, even `yarn.lock` is useless I think.

tkrugg:
@samouss agree, yarn.lock is removed as well in my last commit.
I just read this in the docs:

> Files included with the “package.json#files” field cannot be excluded through .npmignore or .gitignore.

the dist folder is precisely both in .gitignore and in package.json#files. Any recommandation on this?

@Haroenv PR is on its way! https://github.com/algolia/search-insights.js/pull/59

francoischalifour:
@tkrugg There shouldn't be any issue with that. If `dist` is included in the `files` array, it will get picked up by npm (even if part of the ignored files in `.gitignore`).

When you want to make sure that npm picks up the right files, you can run [`npm pack`](https://docs.npmjs.com/cli/pack.html) locally.

Haroenv:
Readme, license, package.json are automatically applied @samouss

francoischalifour:
To learn more about those files, you can refer to the `package.json` [`files`](https://docs.npmjs.com/files/package.json#files) property.

Haroenv:
For the off chance someone might want to import the TS source? Not that important to keep or not keep IMO :)

PR #61, opened 2018-12-14
feat(cookieDuration): add cookieDuration support

tkrugg:
This PR:
- adds cookieDuration param to init
- check it's a valid integer
- and uses it to computed the expire date of the cookie

### DIFF
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -1,11 +1,11 @@
 import AlgoliaInsights from "../insights";
-
-const credentials = {
-  apiKey: "test",
-  applicationID: "test"
-};
+import { userToken } from "../_cookieUtils";
+jest.mock("../_cookieUtils", () => ({ userToken: jest.fn() }));
 
 describe("init", () => {
+  beforeEach(() => {
+    userToken.mockClear();
+  });
   it("should throw if no parameters is passed", () => {
     expect(() => {
       (AlgoliaInsights as any).init();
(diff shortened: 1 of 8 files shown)


lib/__tests__/init.test.ts
```diff
@@ -1,11 +1,11 @@
 import AlgoliaInsights from "../insights";
-
-const credentials = {
-  apiKey: "test",
-  applicationID: "test"
-};
```
tkrugg: was unsused.
I will check tslint config to make sure this doesn't occur more: https://github.com/algolia/search-insights.js/issues/62

lib/_cookieUtils.ts
```diff
@@ -39,29 +40,19 @@ const getCookie = (cname: string): string => {
...
     const newUUID = createUUID();
-    setCookie(COOKIE_KEY, newUUID, 10);
+    setCookie(COOKIE_KEY, newUUID, cookieDuration);
     return newUUID;
   }
-  setCookie(COOKIE_KEY, userToken, 10);
-  return userToken;
-};
 
-/**
```
tkrugg: I moved createUUID outside so I can mock it in the tests. It's generic enough that it can belong inside utils/index.ts

lib/init.ts
```diff
@@ -51,4 +63,10 @@ export function init(options: InitParams) {
 
   // Set hasCredentials
   this._hasCredentials = true;
+
+  const cookieDuration = options.cookieDuration
+    ? options.cookieDuration
+    : 6 * MONTH;
```
tkrugg: by default cookieDuration is 6 months

lib/utils/index.test.ts
```diff
@@ -47,3 +47,9 @@ describe("isFunction", () => {
     expect(isFunction(input)).toEqual(expected);
   });
 });
+
+describe("createUUID", () => {
+  it("should return a string composed of valid hex characters or dashes `-`", () => {
+    expect(createUUID()).toMatch(/^[0-9a-f\-]+$/);
```
tkrugg: No need to test this more. We don't actually care that this is a valid uuid. What's important is that:
- it's random
- it contains only hex characters so that they don't get encoded when used in a cookie.

PR #71, opened 2019-03-01
chore(deps): update stuff

tkrugg:


samouss:
Enable Renovate?

tkrugg:
yep, asked for it to the IT teams, I don't think I'm able to do it by myself on @algolia repos

samouss:
I might be wrong but I think Renovate is enabled on all the Algolia repo. You only have to create the `renovate.json` file. It will then create a PR to setup the configuration.

tkrugg:
Oh didn't know that.
Ok I just added it. Thanks!

PR #99, opened 2019-03-28
feat(version): expose version in main class

tkrugg:
I added this because we need to access current version to send it in the user agent.


lib/insights.ts
```diff
@@ -65,6 +66,8 @@ class AlgoliaAnalytics {
   // LocalStorage
   storageManager: StorageManager;
 
+  version: string = version;
```
tkrugg: should it be VERSION instead?

Haroenv: How do you read this? I think we have mixed between them here and there

scripts/version.sh
```diff
@@ -0,0 +1,11 @@
+# this scripts runs after npm version is ran
+VERSION_FILE="./lib/version.ts";
+DIST_FILE="./dist/search-insights.min.js";
+
+PACKAGE_VERSION=$(cat ./package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[",]//g' | tr -d '[[:space:]]')
```
samouss: Could we dot it in JavaScript? It would be a bit simpler no?

lib/insights.ts
samouss: `version` is fine to me.

scripts/version.sh
tkrugg: I'm later writing in a file, calling git and calling yarn, and those part are simpler (or more readable) in shell to me 😄

lib/insights.ts
tkrugg: > How do you read this? 
You can read this in the console with AlgoliaInsights.version. 
I added it because instantsearch has the same. I don't really need it for the user agent.

scripts/version.sh
samouss: With JavaScript you can call bash command easily like the ones below.

tsconfig.json
```diff
@@ -1,13 +1,12 @@
 {
   "compilerOptions": {
+    "resolveJsonModule": true,
+    "moduleResolution": "node",
     "outDir": "./dist/",
-    "module": "es6",
-    "target": "es6",
+    "module": "commonjs",
+    "target": "es5",
```
tkrugg: I had to change theses following this docs: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html#new---resolvejsonmodule.
I checked the output. I didn't find anything different in the final output. It's just moving the work of the rollup plugin bubble to typescript

lib/insights.ts
```diff
@@ -32,7 +32,12 @@ import {
   InsightsSearchViewFiltersEvent,
   viewedFilters
 } from "./view";
-import { ANONYMOUS_USER_TOKEN, getUserToken, setUserToken } from "./_cookieUtils";
+import {
+  ANONYMOUS_USER_TOKEN,
+  getUserToken,
+  setUserToken
+} from "./_cookieUtils";
+import { version } from "../package.json";
```
samouss: Be careful to not embed the full `package.json` inside the lib.

lib/_sendEvent.ts
```diff
@@ -1,4 +1,5 @@
 import { isNumber, isUndefined, isString, isFunction } from "./utils/index";
+import { version } from "../package.json";
```
yannickcr: Why this import here? (`version` does not seems to be used in this file)

package.json
```diff
@@ -19,7 +19,8 @@
     "build:dev": "rollup --watch --environment NODE_ENV:'development' -c rollup.config.js",
     "build:test": "yarn run jest",
     "build:examples": "webpack --config config/webpack.config.js --color --progress",
-    "dev": "NODE_ENV=development webpack --config config/webpack.config.js --color --progress --watch & node server/server"
+    "dev": "NODE_ENV=development webpack --config config/webpack.config.js --color --progress --watch & node server/server",
+    "version": "yarn build"
```
yannickcr: Do we still need the `yarn version` script?

lib/insights.ts
tkrugg: yes, I was worried about that at first, but rollup-plugin-json is smart about it and even inlines constants

lib/_sendEvent.ts
Haroenv: Since nobody consumes this package from source I think this is good 👍

tsconfig.json
yannickcr: Are you sure about the change in `module` and `target`?

(did not find anything in the doc and it's working on my machine while keeping `es6`)

package.json
tkrugg: yes, I need to rebuild the lib everytime the version changes

lib/_sendEvent.ts
tkrugg: it will be, but it makes no sense at this stage you're right.
Removing

tsconfig.json
tkrugg: yes you're right. I simply followed their doc but that part doesn't seem necessary

PR #101, opened 2019-04-08
chore(StorageManager): remove unused storage manager

tkrugg:
This code was involved in the queryID black magic we decided to get rid
of. It is now no longer necessary.


lib/insights.ts
```diff
@@ -107,16 +103,6 @@ class AlgoliaAnalytics {
   public viewedFilters: (params?: InsightsSearchViewFiltersEvent) => void;
 
   constructor(options?: any) {
-    // Exit on old browsers or if script is not ran in browser
-    if (!document.addEventListener || !window) {
```
samouss: We really want to let the script runs in a different environment than the browser at the moment? IIRC it will throw an error for multiple reasons (XMLHttpRequest, Beacon, Cookie, ...).

tkrugg: It's the plan. It's fairly easy to do and it's what I've been working on this off-sprint.
- Beacon is feature checked before it is used: https://github.com/algolia/search-insights.js/blob/develop/lib/_sendEvent.ts#L124
The goal is to add feature checking for 

- Cookie, to enable react-native compat with the trade-off of no anonymous tracking (started here https://github.com/algolia/search-insights.js/pull/103/files#diff-2bc69471968392e291c0bd99e6836838)
- XMLHttpRequest to enable node compat will come later

Aside from that, there are many things that I find to be wrong with that part of the code.
- both in node and react-native you'd see `document is not defined` rather than the error message, so it doesn't work like it was intended.
- it's doesn't make sense to not work when `addEventListener` is not available because we don't actually use it in the library at all.

samouss: Thanks for the heads up!

PR #102, opened 2019-04-08
fix(typos): various minor unrelated fixes 

tkrugg:
Theses are fixes I've made on the go while working on react-native compat.
They are small enough that I grouped them in a single PR.

### DIFF
--- a/lib/init.ts
+++ b/lib/init.ts
@@ -49,7 +49,7 @@ export function init(options: InitParams) {
       Math.floor(options.cookieDuration) !== options.cookieDuration)
   ) {
     throw new Error(
-      `optional cookieDuration is incorrect, expected an integer`
+      `optional cookieDuration is incorrect, expected an integer.`
     );
   }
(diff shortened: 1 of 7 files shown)


lib/init.ts
```diff
@@ -1,5 +1,4 @@
-import { isUndefined, isString } from "./utils/index";
-import { isNumber } from "util";
```
samouss: Which means that we have `util` in the current build of the lib?

Haroenv: let's add bundlesize to avoid this in the future!

tkrugg: 🤦‍♂️ my mistake.
No it's not in the bundle. rollup doesn't include node builtins unless you tell it to 
https://www.npmjs.com/package/rollup-plugin-node-builtins

tkrugg: Actually you're right. It was included in the bundle 😱 +3kB

PR #103, opened 2019-04-08
fix(compat): disable anonymous tracking on env without cookies

tkrugg:
This disables anonymous tracking on env that do no support cookies.
It is not a breaking change since
- anonymous tracking was broken on theses environements
- environement other than browser were not offcially (ie explicitly) supported

This will allow later usage in react-native and node.


lib/utils/index.ts
```diff
@@ -3,3 +3,10 @@ export const isUndefined = value => typeof value === "undefined";
 export const isString = value => typeof value === "string";
 export const isNumber = value => typeof value === "number";
 export const isFunction = value => typeof value === "function";
+export const supportsCookies = () => {
+  try {
+    return !!navigator.cookieEnabled;
```
Haroenv: not a big deal, but I'd make this
```suggestion
    return Boolean(navigator.cookieEnabled);
```

lib/insights.ts
```diff
@@ -37,6 +37,7 @@ import {
   setUserToken
 } from "./_cookieUtils";
 import { version } from "../package.json";
+import { supportsCookies } from "./utils";
```
samouss: Why do we import it this file since we don't use it?

tkrugg: great catch thanks! fixed here: 77eb366
I found out in the mean time that the rollup typescript plugin we use intentionally doesn't trigger compilation issues. 🙀 

I'll look into moving to https://github.com/ezolenko/rollup-plugin-typescript2.
Gave it a quick try. Quite a few things we've missed 🤦‍♂️

tkrugg:
replaced by https://github.com/algolia/search-insights.js/pull/141

PR #105, opened 2019-04-08
feat(UserAgent): introduce addAlgoliaAgent method

tkrugg:
This introduces a addAlgoliaAgent method that can be also found in the
search client.

```ts
aa('addAlgoliaAgent', 'a random string')
```

### DIFF
--- a/lib/_algoliaAgent.ts
+++ b/lib/_algoliaAgent.ts
@@ -3,8 +3,8 @@ import { version } from "../package.json";
 export const DEFAULT_ALGOLIA_AGENT = `Algolia insights for JavaScript (${version})`;
 
 export function addAlgoliaAgent(algoliaAgent) {
-  if (this._ua.indexOf(`;${algoliaAgent}`) === -1) {
-    this._ua += `;${algoliaAgent}`;
+  if (this._ua.indexOf(`; ${algoliaAgent}`) === -1) {
+    this._ua += `; ${algoliaAgent}`;
     this._uaURIEncoded = encodeURIComponent(this._ua);
   }
 }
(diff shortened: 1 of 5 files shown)


lib/insights.ts
```diff
@@ -66,6 +68,10 @@ class AlgoliaAnalytics {
   _userHasOptedOut: boolean;
   _cookieDuration: number;
 
+  // user agent
+  _ua: string = "";
```
tkrugg: `_ua`  is the current name in the search client too
https://github.com/algolia/algoliasearch-client-javascript/blob/ead86654beb7bc5215bb8ff66d4c951637190a10/src/AlgoliaSearchCore.js#L175

lib/_algoliaAgent.ts
```diff
@@ -0,0 +1,10 @@
+import { version } from "../package.json";
+
+export const DEFAULT_ALGOLIA_AGENT = `Algolia insights for JavaScript (${version})`;
+
+export function addAlgoliaAgent(algoliaAgent) {
+  if (this._ua.indexOf(`;${algoliaAgent}`) === -1) {
+    this._ua += `;${algoliaAgent}`;
```
Haroenv: @samouss separated this with spaces in the client

tkrugg: I look for it as I wasn't sure where the spaces were but couldn't find it.
https://github.com/algolia/algoliasearch-client-javascript/blob/develop/src/AlgoliaSearchCore.js#L173-L177
Am I looking at the right branch?

Haroenv: I think it’s in Robert’s branch still

tkrugg: thanks! it's fixed here: 2546385

PR #107, opened 2019-04-10
feat(getUserToken): add a way to get user token

tkrugg:
As you may know all insights methods are call through an invoker.

**Example:** `aa('init', { appID: "xxx" })` will call `AlgoliaInsights.init({ appID: "xxx" })`.

Yet until now there is no way to get the returned value of a method. It wasn't an issue because most public methods were not returning anything.

However, we want to allow calling `getUserToken` with the invoker and get the returned result (the userToken).
It's useful when you want to be able to call it even when the library is not fully loaded.

The goal for retrieving the user token is usually to add it as a header
on the searchClient. This enables personalized results.

```ts
const searchClient = algoliasearch("APPLICATION_ID", "SEARCH_API_KEY");

// install search insights lib

aa('init', {...})
aa('getUserToken', _, (err, userToken) => {
  searchClient.setExtraHeader("X-Algolia-UserToken", userToken);
})

// search insights loads laters
```

This targets this JIRA ticket https://algolia.atlassian.net/browse/IFW-400


lib/_processQueue.ts
```diff
@@ -14,18 +16,29 @@ export function processQueue(globalObject) {
     const queue = globalObject[pointer].queue || [];
 
     // Loop queue and execute functions in the queue
-    queue.forEach((fn: string[]) => {
-      const functionName = fn[0];
-      const functionArguments = fn[1];
+    queue.forEach((fn: any[]) => {
+      const functionName: string = fn[0];
+      const functionArguments: any = fn[1];
+      const functionCallback: (err: any, res: any) => void = fn[2];
```
tkrugg: we don't actually use `err` for now but it's a pattern enough that I implemented it.
Not very confident about this move. Let me know you thoughts.

README.md
```diff
@@ -196,65 +206,74 @@ To run all examples and play around with the code you have to run two separate c
 ## Migrating from v0 to v1
 
 ### `init` method signature has changed
+
 - `applicationID` is now called `appId`, to stay consistent with our [other js libraries](https://www.algolia.com/doc/guides/building-search-ui/upgrade-guides/js/?language=javascript#previous-usage).
 
 **Before**:
+
```
tkrugg: There are lots of changes right now because I run prettier.
They should be solved when this PR is merged: https://github.com/algolia/search-insights.js/pull/108 and the conflicts solved.

I'll leave this PR as draft until then

lib/_processQueue.ts
Haroenv: why not type this as 

```ts
queue.forEach(
  ([
    functionName: string,
    functionArguments: any,
    functionCallback: (err: any, res: any) => void
  ]) => {}
);
```

francoischalifour: We could even type the queue.

lib/_processQueue.ts
```diff
@@ -14,18 +16,29 @@ export function processQueue(globalObject) {
...
 
-      if (functionName && typeof (this as any)[functionName] === "function") {
-        this[functionName](functionArguments);
+      if (functionName && isFunction((this as any)[functionName])) {
+        const output: any = this[functionName](functionArguments);
+        if (isFunction(functionCallback)) {
+          functionCallback(null, output);
```
francoischalifour: Why don't we pass the error to the callback?

lib/_processQueue.ts
```diff
@@ -14,18 +16,29 @@ export function processQueue(globalObject) {
...
-    globalObject[pointer] = (functionName: string, functionArguments: string) => {
-      (this as any)[functionName](functionArguments);
+    globalObject[pointer] = (
+      functionName: string,
+      functionArguments: string,
+      functionCallback: (err: any, res: any) => void
+    ) => {
+      const output = (this as any)[functionName](functionArguments);
+      if (isFunction(functionCallback)) {
+        functionCallback(null, output);
```
francoischalifour: Why don't we pass the error to the callback?

lib/_processQueue.ts
tkrugg: @Haroenv If you mean
```ts
queue.forEach(
      ([functionName, functionArguments, functionCallback]: [
        string,
        any,
        Function
      ]) => {
      ...
      })
```
It's just that find the current code more legible. 
As françois says we can maybe just type the queue.
```ts
    type QueueItemCallback = (err: any, res: any) => void;
    type QueueItem = [string, any, QueueItemCallback?];
    const queue: QueueItem[] = [];
```

an then maybe just rely on the fact that the rest of the code will infer the typing.
```ts 
queue.forEach(([functionName, functionArguments, functionCallback])
```
WDYT?

lib/_processQueue.ts
tkrugg: There are currently some inconsistencies in a sense that some errors will not fire if comited before processQueue, but do after. 

Theses are inconsistencies I would like to tackle in later PR. They would probably create a big diff. I wanted to keep this PR about "how to get user Token" or more generally "how to get results from a method".

lib/_processQueue.ts
francoischalifour: Yes, it makes more sense to me.

lib/__tests__/_processQueue.test.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+class AlgoliaAnalytics {
+  public init: Function;
+  public otherMethod: Function;
+  public processQueue: Function;
+  constructor() {
+    this.init = jest.fn();
+    this.otherMethod = jest.fn(() => "otherMethodReturnedValue");
+    this.processQueue = processQueue.bind(this);
```
samouss: We could add a comment that explain that it's actually the function that we test.

lib/_processQueue.ts
```diff
@@ -5,27 +5,40 @@
...
-      const functionName = fn[0];
-      const functionArguments = fn[1];
+    queue.forEach((fn: any[]) => {
+      const functionName: string = fn[0];
+      const functionArguments: any = fn[1];
+      const functionCallback: (err: any, res: any) => void = fn[2];
 
-      if (functionName && typeof (this as any)[functionName] === "function") {
-        this[functionName](functionArguments);
+      if (functionName && isFunction((this as any)[functionName])) {
```
samouss: Just by curiosity, is there a reason why skip the function here but not once the queue is proceed?

tkrugg: It's one of the inconsistencies I was thinking of in my answer to François https://github.com/algolia/search-insights.js/pull/107#discussion_r274337122. 
It's been like that for a while. 
I was thinking in a later iteration I'd factorize that code into a single function `processQueueItem` and handle their all theses inconsistencies.

lib/__tests__/_processQueue.test.ts
tkrugg: done here 8798f3e3aebc9b974b132bd430a0f4eb064be39a

README.md
tkrugg: it's done, prettier pr is merged. README.md diff is smaller now.

lib/_processQueue.ts
tkrugg: fixed here: e9646a8

PR #111, opened 2019-04-15
chore(tsc): add type-check for cmdline and CI

tkrugg:
This adds a command that compiles the typescript code and checks the
types.
It is similar to the one that can be found in instantsearch.js:

https://github.com/algolia/instantsearch.js/blame/b3c2154d19bb8e50a07bdd0d5272eb8e3911279e/package.json#L34-L35

**UPDATE:**
pr build is breaking because of a rollup issue I located in the replace plugin.
The plugin has been remove here: https://github.com/algolia/search-insights.js/pull/113
This PR should build when the removal is live on develop.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,7 @@ before_install:
 before_script:
 - yarn install
 - yarn lint
+- yarn type-check
 script:
 - yarn build
 - yarn test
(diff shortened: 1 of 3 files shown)


tkrugg:
@samouss 
> We should enable `strictMode` now that we have a proper compilation. You can opt-out from some of the option independently (e.g. `noImplicitAny`).

Did that here: https://github.com/algolia/search-insights.js/pull/112

PR #115, opened 2019-05-03
fix(processQueue): move callback on method level

tkrugg:
I think this the way it should have been done in the first place.

Queue should pass down args agnostically.
Methods should decide what to do with the args, whether it's a callback or not.

This change results in a better separation of concerns, and will allow implementing callback in other methods without any arity constraint.


lib/_processQueue.ts
```diff
@@ -6,37 +6,28 @@
...
 
   // Check if there is a queue
   if (pointer) {
-    const queue: QueueItem[] = globalObject[pointer].queue || [];
+    const queue: IArguments[] = globalObject[pointer].queue || [];
 
     // Loop queue and execute functions in the queue
-    queue.forEach(([functionName, functionArguments, functionCallback]) => {
+    queue.forEach((args: IArguments) => {
+      const [functionName, ...functionArguments] = [].slice.call(args);
```
yannickcr: I think this could be simplified in:

```ts
queue.forEach(([functionName, ...functionArguments]: IArguments) => {
  if (functionName && isFunction((this as any)[functionName])) {
    this[functionName](...functionArguments);
  }
});
```

lib/_cookieUtils.ts
```diff
@@ -49,6 +50,12 @@ export function setUserToken(userToken: string | number): void {
   }
 }
 
-export function getUserToken() {
+export function getUserToken(
```
yannickcr: Why theses changes to add a callback to `getUserToken`? I have the feeling that it could have been another PR :thinking:

lib/_processQueue.ts
tkrugg: That would have been possible if the queue items were Arrays but they're Arguments instead.
See how typescripts [destructures this](https://www.typescriptlang.org/play/index.html#src=function%20b(%5Bfirst%2C%20...rest%5D)%20%7B%0D%0A%20%20%20%20console.log(first%2C%20rest)%3B%0D%0A%7D%0D%0A%0D%0Ab(2%2C%203)): it's using splice which is not an Argument method.

lib/_cookieUtils.ts
tkrugg: The purpose of the PR is to remove the callback from the queue part and move it on the method level. The only method using the callback right now is `getUserToken`. 

That why I called it, `fix(processQueue): move callback on method level` which is probably a terrible name :)

I can split it in two PR but it breaks the feature continuity insured by this test: https://github.com/algolia/search-insights.js/pull/114/files

lib/_processQueue.ts
yannickcr: Oh, I see, missed that queue was an array of Arguments.

lib/_cookieUtils.ts
yannickcr: Ok if this break a feature then it makes sense to keep both in the same PR :+1:

lib/_processQueue.ts
yannickcr: (But I am very concerned that TypeScript did not trigger any error during my tests)

tkrugg: That's because we compile typescript code to with `target: es6`.
If you change target to `es5` than the error is picked us.

cf. on [typescript playgroup](https://typescript-play.js.org/?allowUnusedLabels=true&downlevelIteration=true&noStrictGenericChecks=true&noUnusedLocals=true&noUnusedParameters=true&esModuleInterop=true&preserveConstEnums=true&removeComments=true&skipLibCheck=true&emitDecoratorMetadata=true&target=2#code/GYVwdgxgLglg9mABAcwBQG1gwE4GcoA0iAdKdgKb4C6AXIgJICC2yIAtuWFLgJSIDeAKEQjEEBLjgAbcsSlw0WPIUQV8PANyCAvoL2hIsBImCo+Q0SlQBDFu07ceOoA) you can try switching target.

I think one way we can  immediately fix this is by changing the target to es5. Compilation can anticipate more errors if make the target as close as possible from the target env.

One thing we can further investigate is getting rid of bublé. Not sure what's the benefit in having it.

yannickcr: Thanks for the explanation. 

This is a bit out of scope with this PR but yes I think it would be nice to fix this so we do not miss some potential errors.

lib/_cookieUtils.ts
```diff
@@ -49,6 +50,12 @@ export function setUserToken(userToken: string | number): void {
...
+  options?: any,
```
samouss: Is there a reason why we expose an `options` argument since we don't use it? To make it easier to extend afterward?

tkrugg: The only reason is: avoid a breaking change.

When the callback was added on `processQueue` level, it assumed all method had the same signature.
`aa('method', {  /* params */ }, function callback(){...})` .
So when you called `getUserToken`, you had to include and uneeded `params` argument
```js
aa('getUserToken', /* options */ null, (err, userToken) => {...})
```

Now that I moved the callback on the method level (which is the way I should have done it in the first place), we could technically get rid of `options` but that would be a breaking change.

lib/_cookieUtils.ts
```diff
@@ -49,6 +50,12 @@ export function setUserToken(userToken: string | number): void {
...
+  callback?: (err: any, userToken: string) => void
+): string {
+  if (isFunction(callback)) {
+    callback(null, this._userToken);
```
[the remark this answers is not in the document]
tkrugg: do you mean having this ?

```ts
if (isFunction(callback)) {
    callback(null, this._userToken);
    return;
}
return this._userToken
```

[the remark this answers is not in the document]
samouss: Yes.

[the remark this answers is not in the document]
tkrugg: I think it's common pattern to return undefined if callback provided or the result otherwise, but I wasn't sure how to type it simply, given the signature is also used here: https://github.com/algolia/search-insights.js/blob/d7599caa40ecfaa3a7904904fd5434e8ee26322e/lib/insights.ts#L93-L96
Then I thought always returning the result was not so bad 😅 and simple too. Any thought?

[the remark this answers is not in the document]
samouss: You have to use the overload, but it's not that important. It's working.

PR #125, opened 2019-06-27
fix(getCookie): support malformed cookie strings

tkrugg:
**Summary** 

This makes the `getCookie()` not fail if the client happens to have malformed cookies we can't decode.
- decode only cookie of interest (cname cookie) rather than the whole
document.cookie string
- wrap decoding a try catch and return empty cookie with a warning.

NB: this is currently going to `0.x` (used on Magento 1 Algolia bundle).
I'll likely have to backport this to the dev branch and release it on `1.x`, although nobody complained about it so far.

### DIFF
--- a/lib/_cookieUtils.ts
+++ b/lib/_cookieUtils.ts
@@ -20,7 +20,7 @@ const setCookie = (cname: string, cvalue: string, exdays: number) => {
  * @param  {[type]} cname [description]
  * @return {[type]}       [description]
  */
-const getCookie = (cname: string):string => {
+export const getCookie = (cname: string):string => {
     var name = cname + "=";
     var decodedCookie = decodeURIComponent(document.cookie);
     var ca = decodedCookie.split(';');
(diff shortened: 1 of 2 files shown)


lib/_cookieUtils.ts
```diff
@@ -20,24 +20,28 @@ const setCookie = (cname: string, cvalue: string, exdays: number) => {
...
+  const ca = document.cookie.split(";");
+  for (let i = 0; i < ca.length; i++) {
+    let c = ca[i];
+    while (c.charAt(0) === " ") {
+      c = c.substring(1);
     }
-    return "";
+    if (c.indexOf(name) === 0) {
+      try {
+        return decodeURIComponent(c.substring(name.length, c.length));
```
Haroenv: decoding isn't necessary here because the only possible values for the uuid are within ascii

tkrugg: good point! fixed heree 44bad6

lib/_cookieUtils.ts
```diff
@@ -20,24 +20,28 @@ const setCookie = (cname: string, cvalue: string, exdays: number) => {
...
+      } catch (e) {
+        console.warn(`Failed to decode ${cname} cookie.`, e);
```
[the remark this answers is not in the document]
tkrugg: yes, actually I removed the test completely, as we no longer decode in e44bad6

PR #132, opened 2019-07-19
chore: add eslint and remove tslint

eunjae-lee:
This PR removes tslint and add eslint instead using `eslint-config-algolia` including typescript support.


eunjae-lee:
@yannickcr Could you take a look at this?
On my mac, `yarn lint` is blocked and never finishes. It's the same even with `--debug` flag.
This isn't urgent at all, though.

yannickcr:
@eunjae-lee I'll have a look

eunjae-lee:
@tkrugg // So many lint errors now. I will do this later when you don't have anything going on in this repo. Otherwise it will make too many conflicts.

PR #133, opened 2019-07-19
chore: update travis config

eunjae-lee:
This PR updates the version of node at travis.


tkrugg:
It looks like you downgrading it. Can you explain the reason?

eunjae-lee:
I'm upgrading from `8.9.0` to `8.16.0` :)
At `.nvmrc`, we're specifying `8.16.0`.

PR #138, opened 2019-07-23
chore: add dotenv

eunjae-lee:
This PR adds `dotenv` to devDependencies.
Instead of running `NODE_ENV=production INDEX_NAME=xxx APP_ID=yyy API_KEY=zzz yarn dev` every time,
we can create a file `.env` and put the following as the content:

```
NODE_ENV=production
INDEX_NAME=xxx
APP_ID=yyy
API_KEY=zzz
```

`.env` is included in `.gitignore`, so won't be commited, but used only locally.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "bundlesize": "0.18.0",
     "cors": "2.8.5",
     "css-loader": "0.28.11",
-    "dotenv": "^8.0.0",
+    "dotenv": "8.0.0",
     "express": "4.17.1",
     "extract-text-webpack-plugin": "3.0.2",
     "html-loader": "0.5.5",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -46,6 +46,7 @@
     "bundlesize": "0.18.0",
     "cors": "2.8.5",
     "css-loader": "0.28.11",
+    "dotenv": "^8.0.0",
```
francoischalifour: This should be pinned.

eunjae-lee: 1f04a79

eunjae-lee:
> We should add a `.env.example` containing the variables without sensitive values. The file is committed but is only a sample that developers need to copy to `.env` to take the variables into account.

A nice suggestion!
f13b65a

PR #141, opened 2019-07-24
fix(compat): node compatibility [PART-2]

eunjae-lee:
## PART-2

This PR disables anonymous tracking on env without cookies.

It's almost copy&paste from https://github.com/algolia/search-insights.js/pull/103

### Stacked PRs
- part1: https://github.com/algolia/search-insights.js/pull/140
- part2: https://github.com/algolia/search-insights.js/pull/141
- part3: https://github.com/algolia/search-insights.js/pull/142


Haroenv:
what's the reason that this is a separate PR from #103?

eunjae-lee:
> what's the reason that this is a separate PR from #103?

No important reason. I wanted to try stacked PRs this time with rollup + cookie + sendBeacon.

lib/init.ts
```diff
@@ -71,4 +71,8 @@ export function init(options: InitParams) {
   // user agent
   this._ua = DEFAULT_ALGOLIA_AGENT;
   this._uaURIEncoded = encodeURIComponent(DEFAULT_ALGOLIA_AGENT);
+
+  if (supportsCookies()) {
+    this.setUserToken(this.ANONYMOUS_USER_TOKEN);
+  }
```
tkrugg: Since we're not automatically calling setUserToken on load, we should throw an error when user has forgotten to set it.

tkrugg: It might be best to delegate that task to sendEvent

eunjae-lee: Oh I see. So we're supposed to throw an error when user hasn't set any user token, for example, in node env?
Setting user token is mandatory before `sendEvent`?

tkrugg: yes there must be no situation where an event can fire with no userToken in the payload.
When the cookie is enabled, this never happens because the anonymous token is always serving as a fallback.
On node, this part is gone. We need to alert the user that they need to either add userToken to their event, or set it with  `setUserToken`

eunjae-lee: https://github.com/algolia/search-insights.js/pull/141/commits/6e0f266cbc56d9df7f641cae832407e79f71f83d

lib/insights.ts
```diff
@@ -141,9 +141,10 @@ class AlgoliaAnalytics {
     this.viewedObjectIDs = viewedObjectIDs.bind(this);
     this.viewedFilters = viewedFilters.bind(this);
 
-    this.setUserToken(this.ANONYMOUS_USER_TOKEN);
     // Process queue upon script execution
-    this.processQueue(window);
+    if (__BROWSER_BUILD__ && typeof window !== "undefined") {
+      this.processQueue(window);
```
samouss: I think that with different entry point for each environment would simplify the logic. We could hook the requester at that point rather than branching inside the code. We could also drop the `__BROWSER_BUILD__` variable. Only the entry point is env specific. The rest of the codebase should be env agnostic (if possible otherwise we keep some feature detection logic like cookies).

lib/_cookieUtils.ts
```diff
@@ -33,7 +33,12 @@ const getCookie = (cname: string): string => {
 export const ANONYMOUS_USER_TOKEN = "ANONYMOUS_USER_TOKEN";
 
 export function setUserToken(userToken: string | number): void {
-  if (userToken === ANONYMOUS_USER_TOKEN) {
+  if (__BROWSER_BUILD__ && userToken === ANONYMOUS_USER_TOKEN) {
```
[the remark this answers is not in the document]
eunjae-lee: right. it's removed https://github.com/algolia/search-insights.js/pull/141/commits/2c5915de7d51edd7f0bf9409945f6793960953eb

lib/insights.ts
eunjae-lee: Now at the entry, it's decided whether or not to call `processQueue`. https://github.com/algolia/search-insights.js/pull/141/commits/baa224fd2f55d5a9e70e158a35eb1a0be1f3fced

lib/entryBrowser.ts
```diff
@@ -1,3 +1,4 @@
 import AlgoliaAnalytics from "./insights";
 
-export default new AlgoliaAnalytics();
+const processQueue = typeof window !== "undefined";
+export default new AlgoliaAnalytics({ processQueue });
```
samouss: I'm not sure it makes sense to have an option dedicated to this behaviour. We could instead run the `processQueue` function only inside the `browser` module. We avoid the introduction of an option only for this purpose, which makes the `insights` a bit more platform agnostic.

eunjae-lee: Okay, since it's executed right after instantiation, we could just call `processQueue` directly to the instance at `browser` module.
👍

eunjae-lee: 5f93971

lib/__tests__/_cookieUtils.test.ts
```diff
@@ -69,10 +81,12 @@ describe("cookieUtils", () => {
       AlgoliaInsights.setUserToken("007");
     });
     it("should return the current userToken", () => {
+      AlgoliaInsights.setUserToken("007");
```
samouss: We already have a `beforeEach` function that calls the function on the line 69.

lib/browser.ts
```diff
@@ -1,3 +1,10 @@
 import AlgoliaAnalytics from "./insights";
+import { processQueue } from "./_processQueue";
 
-export default new AlgoliaAnalytics();
+const instance = new AlgoliaAnalytics({ processQueue });
```
samouss: The argument is not used anymore.

lib/__tests__/harvest.test.ts
```diff
@@ -1,5 +1,5 @@
 import * as testServer from "../../server/server.js";
-import { getInstance } from "../../tests/utils";
+import { getInstanceForBrowser as getInstance } from "../../tests/utils";
```
samouss: Since the function is exactly the same as the `browser` module, why not using the module?

tests/utils.ts
```diff
@@ -1,5 +1,15 @@
 import AlgoliaAnalytics from "../lib/insights";
+import { processQueue } from "../lib/_processQueue";
 
 export function getInstance() {
   return new AlgoliaAnalytics();
 }
+
+export function getInstanceForBrowser() {
```
samouss: https://github.com/algolia/search-insights.js/pull/140#discussion_r309158593

lib/__tests__/_cookieUtils.test.ts
eunjae-lee: https://github.com/algolia/search-insights.js/pull/141/commits/c73dff6202aa65d18dd9e1bc28577c02a9e7d888

lib/browser.ts
eunjae-lee: https://github.com/algolia/search-insights.js/pull/141/commits/b87900fabcca6e7039901ec617032799a93ce765

lib/__tests__/harvest.test.ts
eunjae-lee: It's almost the same. `browser` module exports an instance, and `getInstanceForBrowser` is `a function returning an instance`.
So I've made a `_instance.ts` file to have `getInstanceForBrowser` function and `browser` module now uses this.
The same applies to `node` module.

6f23ec1

lib/__tests__/harvest.test.ts
```diff
@@ -1,5 +1,5 @@
 import * as testServer from "../../server/server.js";
-import { getInstance } from "../../tests/utils";
+import { getInstance } from "../_instance";
```
Haroenv: why  is this moved again here?

eunjae-lee: `browser` and `node` modules were exporting an instance.
And we need "a function" that returns an instance in test cases.
So I had to have the functions in `_instance` file, so that I can use them in test cases
and also in `browser` and `node` module.

lib/insights.ts
```diff
@@ -113,9 +112,8 @@ class AlgoliaAnalytics {
   public viewedObjectIDs: (params?: InsightsSearchViewObjectIDsEvent) => void;
   public viewedFilters: (params?: InsightsSearchViewFiltersEvent) => void;
 
-  constructor(options?: any) {
+  constructor({  }: AlgoliaAnalyticsOptions = {}) {
```
tkrugg: is there a reason for changing in this PR?

eunjae-lee: Actually it's a stepping stone for this: https://github.com/algolia/search-insights.js/pull/142/files#diff-cc2c0a8a8dfcd80e79e33bc5d480b745R119
I should've left a comment here to avoid confusion.

lib/_instance.ts
```diff
@@ -0,0 +1,22 @@
...
+    processQueue.call(instance, window);
+  }
+  return instance;
+}
+
+export function getInstanceForNode() {
+  const instance = new AlgoliaAnalytics();
+  const functionalInterface = getFunctionalInterface(instance);
+  return functionalInterface;
+}
```
tkrugg: Wouldn't it be simpler if we just put getInstantForNode inside node
and getInstanceForBrowser inside browser.ts ?

lib/_sendEvent.ts
```diff
@@ -33,6 +33,11 @@ export function sendEvent(
       "Before calling any methods on the analytics, you first need to call the 'init' function with appId and apiKey parameters"
     );
   }
+  if (!this._userToken) {
+    throw new Error(
+      "Before calling any methods on the analytics, you first need to call 'setUserToken' function."
+    );
+  }
```
tkrugg: 👍 Awesome! this is the best place to put it!

lib/_instance.ts
eunjae-lee: I tried put `getInstanceForBrowser` inside `browser.ts`. In the `browser.ts`, there were a default export `getInstanceForBrowser()` and a named export `getInstanceForBrowser`.
And it somehow broke the umd build. So I had to get rid of the named export.

PR #142, opened 2019-07-25
fix(compat): node compatibility [PART-3]

eunjae-lee:
## PART-3

This PR provides fallbacks to `sendBeacon` function.
It tries in this order:
1. navigator.sendBeacon
2. XMLHttpRequest
3. "http" or "https" from node

### Stacked PRs
- part1: https://github.com/algolia/search-insights.js/pull/140
- part2: https://github.com/algolia/search-insights.js/pull/141
- part3: https://github.com/algolia/search-insights.js/pull/142


lib/utils/__tests__/featureDetection.test.ts
```diff
@@ -1,9 +1,40 @@
...
+describe("featureDetection in jsdom env", () => {
   describe("supportsCookies", () => {
-    it("returns true in jsdom env", () => {
+    it("returns true", () => {
       expect(supportsCookies()).toBe(true);
     });
   });
+
+  describe("supportsSendBeacon", () => {
+    it("should return false if non available", () => {
```
tkrugg:
```suggestion
    it("should return false if not available", () => {
```

lib/utils/__tests__/request.test.ts
```diff
@@ -0,0 +1,132 @@
...
+      headers: {
+        "Content-Type": "application/json",
+        "Content-Length": JSON.stringify(data).length
+      }
+    });
+    expect(write).toHaveBeenCalledTimes(1);
+    expect(write).toHaveBeenLastCalledWith(JSON.stringify(data));
+  });
+
+  it("should send with nodeHttpsRequest", () => {
```
tkrugg:
```suggestion
  it("should send with nodeHttpsRequest if url starts with https://", () => {
```

lib/utils/__tests__/request.test.ts
```diff
@@ -0,0 +1,132 @@
...
+    expect(navigator.sendBeacon).not.toHaveBeenCalled();
+    expect(open).toHaveBeenCalledTimes(1);
+    expect(open).toHaveBeenLastCalledWith("POST", url);
+    expect(send).toHaveBeenCalledTimes(1);
+    expect(send).toHaveBeenLastCalledWith(JSON.stringify(data));
+    expect(nodeHttpRequest).not.toHaveBeenCalled();
+    expect(nodeHttpsRequest).not.toHaveBeenCalled();
+  });
+
+  it("should send with nodeHttpRequest", () => {
```
tkrugg:
```suggestion
  it("should send with nodeHttpRequest if url does not start with https://", () => {
```

eunjae-lee:
> Actually let's remove that queue processing mechanism altogether on node, I can't think of any use case on node environment, and none of the other server side clients have it.

Then, I will add a function `supportsWindow` and wrap `processQueue()` with it.
At build time, I cannot remove the code. Because `cjs` build also can run on browser.
What do you think?

eunjae-lee:
I should add more test with `@jest-environment node`, to check from loading the library to sending events in node env.

lib/utils/featureDetection.ts
```diff
@@ -1,7 +1,34 @@
+import { request as nodeHttpRequest } from "http";
+import { request as nodeHttpsRequest } from "https";
```
Haroenv: I think this will cause a polyfill for the node default modules in the regular bundle, unless you specifically exclude them in rollup

lib/utils/request.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (supportsSendBeacon()) {
+    return requestWithSendBeacon;
+  }
+
+  if (supportsXMLHttpRequest()) {
+    return requestWithXMLHttpRequest;
+  }
+
+  if (supportsNodeHttpModule()) {
+    return requestWithNodeHttpModule;
```
Haroenv: if we dynamically require('https') here, I think there might be less of an effect of having the module

lib/utils/featureDetection.ts
Haroenv: looks like you are doing that

lib/utils/request.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+const requestWithXMLHttpRequest: RequestType = (url, data) => {
+  const serializedData = JSON.stringify(data);
+  const report = new XMLHttpRequest();
+  report.open("POST", url);
+  report.send(serializedData);
+};
+
+const requestWithNodeHttpModule: RequestType = (url, data) => {
```
samouss: If we go with two entry point we could link the requester to each build rather than have each requester inside the code. Then we can even imagine exposing an option to provide your own requester at some point.

We'll have to feature detect some of the features still (like `beacon` for example).

eunjae-lee: @samouss Thanks! 
I was struggling to deal with both detecting features and __BROWSER_BUILD__.
Having two entry point makes things much simpler.

eunjae-lee: Now at the entry, `requestFn` is provided to insights class constructor. https://github.com/algolia/search-insights.js/pull/142/commits/2a5cc21db14ab99f33c6bb79746a9a05e5d3bdd7

lib/__tests__/_sendEvent.node.test.ts
```diff
@@ -1,23 +1,23 @@
...
-    AlgoliaInsights = getInstance();
-    AlgoliaInsights.init(credentials);
+    aa = getInstance();
+    aa("init", credentials);
   });
 
   it("throws when user token is not set", () => {
     expect(() => {
-      (AlgoliaInsights as any).sendEvent("click", {
+      aa("sendEvent", "click", {
```
tkrugg: I see you have migrated this one to the exposed API but not the other ones. Is there reason for not having all tests using the same API ?

eunjae-lee:
I admit these stacked PRs are messy and I shouldn't have done this.
I hope to make PRs cleaner next time.
After a discussion with Youcef on Wednesday, I've made some changes.
Some of the changes don't belong to this PR, but please bear with me for this PR this time.

- https://github.com/algolia/search-insights.js/pull/142/commits/3f7f2a1f1eabf84a337c8fd50003b7666522fc49: I've removed `_instance` and the most test cases use `AlgoliaAnalytics` directly now. It turned out that the most test cases don't need the extra things that `getInstance()` functions did, for example, using real requestFn and processQueue.
- https://github.com/algolia/search-insights.js/pull/142/commits/6af42999ac774a07f73e590f8ae2f9f58e35eac3: `request.ts` file had (A) request functions and (B) requester functions returning request function. I've splitted them and now there're `getRequesterForBrowser` and `getRequesterForNode`. 
- https://github.com/algolia/search-insights.js/pull/142/commits/600110089db0dd519b5df64ddf7f77e662cc86cc: I've found out that Node 8.x and Node 10.x have different signatures for `http/https.request()` function. This commit changes the parameter to support both versions.
    - node 8.x: https://nodejs.org/docs/latest-v8.x/api/https.html#https_https_request_options_callback
    - node 10.x: https://nodejs.org/dist/latest-v10.x/docs/api/https.html#https_https_request_options_callback
- https://github.com/algolia/search-insights.js/pull/142/commits/ad3fe01106e5f6eb335a1a3efb06efd9655b236e: This commit allows user to put `userToken` inside event payload. 

---

To test if this works in node,
```bash
$ yarn pack
$ cd ~/sandbox
$ mkdir test && cd test
$ yarn add <search-insights-path>/search-insights-v1.2.1.tgz
$ node
```

```js
aa = require('search-insights')
aa('init', {appId: 'a', apiKey: 'b'})
aa('clickedObjectIDs', {index: 'a', eventName: 'add to basket', objectIDs: []}) // <- SHOULD FAIL
aa('clickedObjectIDs', {index: 'a', eventName: 'add to basket', objectIDs: [], userToken: 'a'}) <- SHOULD SUCCEED
```

lib/__tests__/_sendEvent.test.ts
```diff
@@ -31,8 +37,15 @@ describe("sendEvent", () => {
   });
 
   describe("with XMLHttpRequest", () => {
+    let AlgoliaInsights;
+    let sendBeaconBackup = window.navigator.sendBeacon;
```
tkrugg: I guess you don't have to intialize it here if it's done in the beforeEach

eunjae-lee: right. I've fixed! https://github.com/algolia/search-insights.js/pull/142/commits/faa50f7cd8420c8979547bedfe07fefa9dd6eec5

PR #143, opened 2019-07-25
chore: add jest watch typeahead plugin

eunjae-lee:
This PR adds [typeahead](https://github.com/jest-community/jest-watch-typeahead) plugin to jest.

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "instantsearch.js": "2.10.5",
     "jest": "24.8.0",
     "jest-localstorage-mock": "2.4.0",
+    "jest-watch-typeahead": "^0.3.1",
     "lunr": "2.3.6",
     "node-sass": "4.12.0",
     "prettier": "1.18.2",
@@ -95,6 +96,10 @@
       "js",
       "jsx",
       "json"
+    ],
+    "watchPlugins": [
+      "jest-watch-typeahead/filename",
+      "jest-watch-typeahead/testname"
     ]
   },
   "dependencies": {},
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -55,6 +55,7 @@
     "instantsearch.js": "2.10.5",
     "jest": "24.8.0",
     "jest-localstorage-mock": "2.4.0",
+    "jest-watch-typeahead": "^0.3.1",
```
francoischalifour: This should be pinned.

eunjae-lee: I'm not used to it, so I keep forgetting it 😅 Thanks for pointing it out!
Btw, is this guide aligned to what you do?

1. Any apps (web or node.js) that aren't require()'d by other packages should pin all types of dependencies for greatest reliability/predictability.
2. Browser or dual browser/node.js libraries that are consumed/required()'d by others should keep using semver ranges for dependencies but can use pinned dependencies for devDependencies.
3. Node.js-only libraries can consider pinning all dependencies, because application size/duplicate dependencies are not as much a concern in node.js compared to the browser. Of course, don't do that if your library is a micro one likely to be consumed in disk-sensitive environments.

from [here](https://renovatebot.com/docs/dependency-pinning/)

eunjae-lee:
> Thanks @eunjae-lee
> As françois mentionned, it's best to pin the dependency for consistency but otherwise, cool addition!

Thanks! Just did ;)

package.json
francoischalifour: Yep! We follow these rules in our [Renovate configuration](https://github.com/algolia/renovate-config-algolia). It's better to always pin dev dependencies so that every collaborator has exactly the same environment.

eunjae-lee: Thanks for the confirmation!
I wish to have some repo-level configuration which automatically adds `--exact` when I add devDepencency.
I googled but couldn't find any. I will try not to forget :)

PR #183, opened 2019-12-05
test(setUserToken): check setUserToken and cookie behaviour

tkrugg:
This PR adds e2e tests for `setUserToken` and cookie. It is not adding any feature.

### DIFF
--- a/lib/__tests__/harvest.test.ts
+++ b/lib/__tests__/harvest.test.ts
@@ -137,7 +137,7 @@ describe("Integration tests", () => {
 
         const cookies = await page.cookies();
         const algoliaCookie = cookies.find(
-          cookie => (cookie.name = "_ALGOLIA")
+          cookie => cookie.name === "_ALGOLIA"
         );
 
         expect(userToken).toMatch(/^anonymous-[-\w]+$/);
--- a/lib/__tests__/harvest.test.ts
+++ b/lib/__tests__/harvest.test.ts
@@ -124,16 +124,38 @@ describe("Integration tests", () => {
       it("should retrieve a queryID on page load", async () => {
         expect(data).toHaveProperty("queryID");
       });
-      it("should generate an anonymous userToken on init", async () => {
-        const userToken = await page.evaluate(() => {
-          return new Promise((resolve, reject) =>
-            window.aa("getUserToken", null, (err, res) => {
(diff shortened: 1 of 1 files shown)


lib/__tests__/harvest.test.ts
```diff
@@ -124,16 +124,38 @@ describe("Integration tests", () => {
...
+              window.aa("getUserToken", null, (err, res) => {
+                if (err) return reject(err);
+                resolve(res);
+              })
+            )
+        );
+
+        const cookies = await page.cookies();
+        const algoliaCookie = cookies.find(
+          cookie => (cookie.name = "_ALGOLIA")
```
francoischalifour: Not sure to understand this assignment. Did you mean this?

```suggestion
          cookie => cookie.name === "_ALGOLIA"
```

tkrugg: yes absolutely, thanks for catching the typo

lib/__tests__/harvest.test.ts
```diff
@@ -124,16 +124,38 @@ describe("Integration tests", () => {
...
+              window.aa("getUserToken", null, (err, res) => {
+                if (err) return reject(err);
+                resolve(res);
+              })
+            )
+        );
+
+        const cookies = await page.cookies();
+        const algoliaCookie = cookies.find(
+          cookie => cookie.name === "_ALGOLIA"
```
francoischalifour: Unrelated to this PR but do you have context about why the cookie that stores the user token is called `_ALGOLIA`?

tkrugg: no, seems like it was there from day one.
https://github.com/algolia/search-insights.js/commit/8cac6358ac667e06933bb8a10c03e43250cccc13#diff-2bc69471968392e291c0bd99e6836838R3
Do you find the name confusing?

francoischalifour: Since it's a generic name I expected it to store an object, not only the user token string. But there might be a reason.

tkrugg: I agree. I think it was inspired by [Google Analytics' `_ga`](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#cookieName).

PR #184, opened 2019-12-11
feat: add alg_user_token cookie for storing set userToken

tkrugg:


lib/_cookieUtils.ts
```diff
@@ -1,7 +1,10 @@
 import { createUUID } from "./utils/uuid";
 import { isFunction, supportsCookies } from "./utils";
 
-const COOKIE_KEY = "_ALGOLIA";
+const USER_TOKEN_COOKIE_KEY = "alg_user_token";
+const ANONYMOUS_TOKEN_COOKIE_KEY = "_ALGOLIA";
+// TODO: rename this to alg_anonymous_token in a future release
+// Add code in init checking if _ALOGLIA cookie exists and replaces it in alg_anonymous_token
```
Haroenv:
```suggestion
// Add code in init checking if _ALGOLIA cookie exists and replaces it in alg_anonymous_token
```

Haroenv:
just a wild question, but why not adding the non-generated user token to the _ALGOLIA cookie?

tkrugg:
> just a wild question, but why not adding the non-generated user token to the _ALGOLIA cookie?

I had a first version which did just that but it had two downside:
- it needed more "parsing" 
- the anonymous token and the user token had to share the same expiration date, yet there were not set at the same time.

Separating it appeared more simple

francoischalifour:
What are the implications of this change?

Not sure if relevant but "alg" can mean "algorithm".

tkrugg:
there are @francoischalifour no implication.

Haroenv:
I think what Francois is saying that maybe another service/cookie could pick this name

francoischalifour:
It's hard to understand where you're going without any context. There is no description provided in the PR message.

It would be nice that you write a summary that answers some of the following questions:

- Why this change? What are you trying to achieve?
- Are there any concerns to have regarding backward compatibility?
- Can this namespace already be taken?
- Why have another cookie?
- Why such a different name from `_ALGOLIA`?

Those are some of the questions that when answered in the PR message, would help us better understand what's going on and help us give proper feedback when it's ready.

Even tough it's a draft, you proposed algolia/instantsearch.js#4265 which relies on it – it would be helpful to give more context to this PR :)

tkrugg:
Hi @francoischalifour please note this is a draft PR.
I'll provide a description and hopefully answer all your concerns by the time it's open to review.

PR #187, opened 2019-12-20
chore: fix the insights snippet in readme.md

tkrugg:
- make the globals name configurable (using parameter `s` instead of `.aa`)
- pass source URL from parameters
- remove protocol from source url
- update to the latest version 1.3.1

This is not breaking change.


README.md
```diff
@@ -37,12 +37,12 @@ We recommend loading the library by adding the snippet below to all pages where
 <!-- prettier-ignore-start -->
 ```html
 <script>
-var ALGOLIA_INSIGHTS_SRC = "https://cdn.jsdelivr.net/npm/search-insights@1.3.0";
+var ALGOLIA_INSIGHTS_SRC = "//cdn.jsdelivr.net/npm/search-insights@1.3.1";
```
Haroenv: I'd leave https here

tkrugg: yes, https://github.com/algolia/doc/pull/4016

eunjae-lee: For just curiosity, what happens http version of that is loaded on a http website?
any difference between that and loading https version on a http website?

Haroenv: on a http website, this will load the http version of the script, which used to be common place when older IE would not load a  mixed (https on http) script, however that's no longer the case, and thus only makes the link less secure

eunjae-lee: Oh, I thought `http on https` is the only mixed script.
If that's no longer the case, it makes sense to keep `https:` 👍

eunjae-lee: wdyt @tkrugg ?

tkrugg: I guess yes. I’ll make a PR on the docs so that it’s the same everywhere

tkrugg: restored in last commit.
Also opened a PR on the docs: https://github.com/algolia/doc/pull/4062

PR #188, opened 2019-12-24
chore: update the version in the snippet at release

eunjae-lee:
This PR will
- update the version in the snippet of README.md before release
- update the version of ship.js.


ship.config.js
```diff
@@ -6,5 +9,13 @@ module.exports = {
     }
   },
   buildCommand: () => "yarn build && /bin/bash ./pre-deploy.sh",
-  pullRequestReviewer: "@algolia/instantsearch-for-websites"
+  pullRequestReviewer: ["@algolia/instantsearch-for-websites"],
+  versionUpdated: ({ version, dir }) => {
+    const readmePath = path.resolve(dir, "README.md");
+    let content = fs.readFileSync(readmePath).toString();
+    const regex = /cdn\.jsdelivr\.net\/npm\/search-insights@(\d+?\.\d+?\.\d+?)/;
```
Haroenv: could the version not also have text (like `alpha`)?

francoischalifour: It could, but we probably won't mention installing an `alpha` version from the README.

ship.config.js
```diff
@@ -6,5 +9,13 @@ module.exports = {
     }
   },
   buildCommand: () => "yarn build && /bin/bash ./pre-deploy.sh",
-  pullRequestReviewer: "@algolia/instantsearch-for-websites"
+  pullRequestReviewer: ["@algolia/instantsearch-for-websites"],
+  versionUpdated: ({ version, dir }) => {
```
francoischalifour: Do we want to call this function only if it's a stable version? We might not want to update the README if we release an `alpha` version.

yannickcr: Same kind of question, how does ship.js handle pre-release versions? (from the documentation the `version` format is in `x.y.z` but do not precise versions with additional metadata like `1.0.0-alpha`)

eunjae-lee: @Haroenv @francoischalifour good point!
`versionUpdated` hook is called with `version, dir, releaseType`.
So I will update it to check first if `releaseType in ['major', 'minor', 'patch']`.

@yannickcr 
For example, the current version is `1.2.3`.
You want to release `1.2.4-alpha.0`.
When you run `yarn release:prepare`, it will ask you if x.y.z is the correct next version.
Then you'll say "no", and you'll manually type `1.2.4-alpha.0`.
Then it proceeds.

And when it publishes to NPM, it will automatically attach `--tag alpha` to make sure it's released as an alpha version.

Later, you run again `yarn release:prepare`, ship.js will ask you if `1.2.4-alpha.1` is the correct next version. Now you want to just release `1.2.4`. Then you'll say "no" and manually type `1.2.4`.
Then it proceeds. And it publishes to NPM with `--tag latest`.

Does this answer your question? Let me know if there's anything unclear!

Haroenv: not sure if you did it for brevity, but `in` wouldn't be valid there, and it should be: `['major', 'minor', 'patch'].includes(releaseType)`

yannickcr: @eunjae-lee my question was more what are the values of `version` and `releaseType` in the case of a prerelease version?

eunjae-lee: > not sure if you did it for brevity, but `in` wouldn't be valid there, and it should be: `['major', 'minor', 'patch'].includes(releaseType)`

@Haroenv yes, I was going for pseudo code, but I realize the length isn't different much 😂 

@yannickcr ah okay, good 🙂

PR #189, opened 2020-01-10
chore(release): update merge strategy

eunjae-lee:


ship.config.js
```diff
@@ -3,10 +3,7 @@ const path = require("path");
 
 module.exports = {
   mergeStrategy: {
-    toSameBranch: ["maint"],
-    toReleaseBranch: {
-      develop: "master"
-    }
+    toSameBranch: ["maint", "master"]
```
Haroenv: what does maint mean? Is this `maintenance`? If so I suggest making the name full

francoischalifour: That's the name of the [maintenance branch](https://github.com/algolia/search-insights.js/tree/maint). We should even call it `v0` IMO.

eunjae-lee: Now that I'm thinking about it, I'm not sure if we need to maintain v0..?

tkrugg: we have several customers using it (many shopify and magento), and the migration from v0 to v1 is not trivial (because conversion events require the queryID in v1).

Last time we release 0.x was 6 months ago, so having a way to release the `maint` branch would be good, especially since we release only every one in a while, IMHO.

eunjae-lee: Okay, then we keep v0. I will create a branch named `v0` instead of `maint`. 
And I realize I need to create another PR to `v0` to add Ship.js to that branch because this PR adds Ship.js to only `master` branch.

PR #196, opened 2020-01-29
chore: remove NODE_ENV from .env.example

eunjae-lee:
This PR removes unnecessary `NODE_ENV=production` from `.env.example`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,4 +1,3 @@
-NODE_ENV=production
 INDEX_NAME=instant_search
 APP_ID=latency
 API_KEY=xxx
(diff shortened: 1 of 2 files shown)


Haroenv:
I don't see how the test failure can be related, but can you investigate it?

eunjae-lee:
> I don't see how the test failure can be related, but can you investigate it?

yup, on it !

eunjae-lee:
If you run `yarn run build:examples`, it will fail with an error message that it requires NODE_ENV.
I guess it's okay to leave it as is. We don't have to update the `build:examples` script to include `NODE_ENV=production`.

PR #200, opened 2020-02-13
feat(sendEvent): return the value of the request function

### LINKED ISSUE #199: Add wait on response
I'd like to do the following in the browser:

```
const res = await aa.sendEvent(this.eventType, this.eventJson);
console.log(res);
```

For that to be possible I think we would need to both:
- allow to force the requester (xml)
- return a promise

Haroenv:
The existing request functions return undefined, so this isn't a breaking change, but this allows you to have a custom requester you can wait on (e.g. fetch).

fixes #199

### DIFF
--- a/lib/__tests__/_sendEvent.test.ts
+++ b/lib/__tests__/_sendEvent.test.ts
@@ -12,10 +12,8 @@ const credentials = {
   appId: "testId"
 };
 
-function setupInstance() {
-  const instance = new AlgoliaAnalytics({
-    requestFn: getRequesterForBrowser()
-  });
+function setupInstance(requestFn = getRequesterForBrowser()) {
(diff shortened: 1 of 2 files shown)


Haroenv:
Don’t need to change the implementations, since that’s a breaking change, but this is for a custom requestFn to which you return something. Can add a test

lib/__tests__/_sendEvent.test.ts
```diff
@@ -154,6 +152,50 @@ describe("sendEvent", () => {
     });
   });
 
+  describe("with custom requestFn", () => {
+    let analyticsInstance;
+    const fakeRequestFn = jest.fn();
```
tkrugg: shouldn't this be reinitialized in the beforeEach?

PR #203, opened 2020-06-24
feat: export `createInsightsClient` for node env

eunjae-lee:
This PR exports `createInsightsClient` for node env. This way, users can create a client with their own requester function. For example, in native react environment, the default http requester doesn't work. Instead of providing all possible requesters, we can expose this API to support more environments and use cases.


lib/node.ts
```diff
@@ -2,8 +2,12 @@ import AlgoliaAnalytics from "./insights";
...
 import { getRequesterForNode } from "./utils/getRequesterForNode";
 
-const requestFn = getRequesterForNode();
-const instance = new AlgoliaAnalytics({ requestFn });
-const functionalInterface = getFunctionalInterface(instance);
+export function createInsightsClient(requestFn) {
+  const instance = new AlgoliaAnalytics({ requestFn });
+  const functionalInterface = getFunctionalInterface(instance);
+  return functionalInterface;
+}
```
Haroenv, line 9:
```suggestion
export function createInsightsClient(requestFn) {
  return getFunctionalInterface(new AlgoliaAnalytics({ requestFn }));
}
```

lib/node.ts
```diff
@@ -2,8 +2,12 @@ import AlgoliaAnalytics from "./insights";
...
 
-export default functionalInterface;
+const requestFn = getRequesterForNode();
+const insightsClient = createInsightsClient(requestFn);
+export default insightsClient;
```
Haroenv, line 13:
```suggestion
export default createInsightsClient(getRequesterForNode());
```

PR #206, opened 2020-07-01
Fix types

adambullmer:
- Adds typescript definition output
- Fixes missing required eventName interface property

_sendEvent identifies the eventName property as optional, and then does a type assertion on it in code, therefore making it required, even if empty. Requiring it on all higher level interfaces to avoid these type checking issues


adambullmer:
Needs some more throught before a PR is made on how this library is intended to be consumed. Unfortunately including TS files forces them into the consuming repo's linting standards, which is a problem for anyone with strict TS standards.

PR #207, opened 2020-07-01
fix(types): add definition output and require eventName

adambullmer:
- Adds typescript definition output
- Fixes missing required eventName interface property

_sendEvent identifies the eventName property as optional, and then does a type assertion on it in code, therefore making it required, even if empty. Requiring it on all higher level interfaces to avoid these type checking issues

This should also address the concerns of #9

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 dist/
 types/
+lib/**/*.js
 node_modules
 tests/screenshots
 tests/production
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,4 @@
 dist/
-types/
-lib/**/*.js
 node_modules
 tests/screenshots
 tests/production
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 dist/
+types/
 node_modules
 tests/screenshots
 tests/production
--- a/package.json
+++ b/package.json
@@ -4,14 +4,13 @@
   "version": "1.4.0",
   "jsdelivr": "dist/search-insights.min.js",
   "main": "dist/search-insights.cjs.min.js",
-  "types": "types/lib/node.d.ts",
+  "types": "dist/lib/node.d.ts",
   "engines": {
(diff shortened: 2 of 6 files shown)


adambullmer:
Confirmed that using unbundled code will retain emitted type definitions, and stop throwing errors in client build processes for differing strictness opinions

.gitignore
```diff
@@ -1,6 +1,8 @@
 dist/
+types/
+lib/**/*.js
```
eunjae-lee, line 3: @adambullmer Can you explain what this line is for?

package.json
```diff
@@ -4,18 +4,21 @@
...
   "files": [
     "dist",
     "lib",
+    "types",
     "!**/__tests__/**"
   ],
   "scripts": {
-    "build": "rollup --environment NODE_ENV:'production' -c rollup.config.js",
-    "build:dev": "rollup --watch --environment NODE_ENV:'development' -c rollup.config.js",
+    "build": "yarn build:types && rollup --environment NODE_ENV:'production' -c rollup.config.js",
```
eunjae-lee, line 18: @adambullmer I tried `yarn run build` on my computer, and it didn't generate `types` folder. Can you look into this?

.gitignore
adambullmer: That's actually from some rapid iteration on getting this feature to work properly and no longer necessary. I'll remove to reduce confusion and unintended file ignoring

package.json
adambullmer: This will actually output the types into the dist directory, colocated with their corresponding JS files. This was necessary to fix relative linking in the definition outputs, and for definition resolution.

etienne-martin:
Hey @adambullmer you might want to check this PR: https://github.com/algolia/search-insights.js/pull/243

Haroenv:
We have moved with the subsequent PR, since it required less changes. Thanks for this pull request though @adambullmer !

PR #209, opened 2020-07-15
feat: add onUserTokenChange to know when token is assigned

eunjae-lee:
This PR adds `onUserTokenChange` so that you can attach a callback to receive user token changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,32 @@ aa('getUserToken', null, (err, userToken) => {
 });
 ```
 
+#### Listen to `userToken` change
+
+If you want to attach a listener for `userToken` change, you can call `onUserTokenChange`.
+
+```js
+aa('onUserTokenChange', (userToken) => {
+  console.log("userToken has changed: ", userToken);
+});
+```
+
+`onUserTokenChange` accepts `callback`(required) and `options`(optional).
+
+```js
+aa('onUserTokenChange', callback, options);
+```
+
+| Option      | Type       | Description                                    |
+| ----------- | ---------- | ---------------------------------------------- |
+| `immediate` | `boolean`  | Fire the callback as soon as it's attached     |
+
+If there is a chance that userToken is already set before `onUserTokenChange`, then you can provide `{ immediate: true }` option in order to get the token as soon as you attach `onUserTokenChange`.
+
+```js
+aa('onUserTokenChange', callback, { immediate: true });
+```
+
 #### Report a click event
 
 ```js
--- a/README.md
+++ b/README.md
@@ -219,11 +219,24 @@ aa('onUserTokenChange', callback, options);
 | ----------- | ---------- | ---------------------------------------------- |
 | `immediate` | `boolean`  | Fire the callback as soon as it's attached     |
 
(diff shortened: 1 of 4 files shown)


eunjae-lee:
For some reason I don't know, the status is not being updated from Travis while the task was successfully done over there.

To make sure, I ran `yarn lint && yarn test && yarn type-check` locally and it succeeded.

lib/__tests__/init.test.ts
```diff
@@ -135,4 +135,42 @@ describe("init", () => {
...
+      analyticsInstance.init({ apiKey: "***", appId: "XXX", region: "de" });
+      // Because cookie is enabled, anonymous token must be generated already.
+      expect(analyticsInstance._userToken).toBeTruthy();
+      expect(analyticsInstance._userToken.length).toBeGreaterThan(0);
+      const callback = jest.fn();
+      analyticsInstance.onUserTokenChange(callback, { immediate: true });
+      expect(callback).toHaveBeenCalledWith(analyticsInstance._userToken);
+      expect(callback).toHaveBeenCalledTimes(1);
+
+      analyticsInstance.setUserToken("abc");
```
Haroenv, line 154: can you add a test for setUserToken to be called first?

README.md
```diff
@@ -199,6 +199,32 @@ aa('getUserToken', null, (err, userToken) => {
...
+
+```js
+aa('onUserTokenChange', callback, options);
+```
+
+| Option      | Type       | Description                                    |
+| ----------- | ---------- | ---------------------------------------------- |
+| `immediate` | `boolean`  | Fire the callback as soon as it's attached     |
+
+If there is a chance that userToken is already set before `onUserTokenChange`, then you can provide `{ immediate: true }` option in order to get the token as soon as you attach `onUserTokenChange`.
```
francoischalifour, line 222: Instead of "If there is a chance that [...]", can we give an example when this can happen? It might get even more confusing for users otherwise.

lib/__tests__/init.test.ts
```diff
@@ -135,4 +135,42 @@ describe("init", () => {
     setUserToken.mockRestore();
     supportsCookies.mockRestore();
   });
+
+  describe("callback for userToken", () => {
```
francoischalifour, line 139: We never test `onUserTokenChange` without the `immediate` option.

README.md
eunjae-lee: thanks. wdyt? https://github.com/algolia/search-insights.js/pull/209/commits/76304708bf4a800fdef4e34327b25b5eeae39f37

lib/__tests__/init.test.ts
eunjae-lee: https://github.com/algolia/search-insights.js/pull/209/commits/f8d2c56c58fafa7eb6cd14885404a6b34917e7d5

lib/__tests__/init.test.ts
eunjae-lee: https://github.com/algolia/search-insights.js/pull/209/commits/f8d2c56c58fafa7eb6cd14885404a6b34917e7d5

PR #210, opened 2020-07-15
fix: return value from getUserToken

eunjae-lee:
This PR fixes the regression where it didn't return value from `getUserToken` correctly when it was used without callback.

Usage1. `aa("getUserToken", options, callback)` <- worked

Usage2. `const token = aa("getUserToken")` <- didn't worked. Fixed in this PR.

A working example: https://codesandbox.io/s/vanilla-dm9d7?file=/src/index.js

### DIFF
--- a/lib/__tests__/node.test.ts
+++ b/lib/__tests__/node.test.ts
@@ -1,0 +1,15 @@
+import insightsClient from "../node";
+
+describe("node", () => {
+  it("correctly return userToken", done => {
+    insightsClient("init", { appId: "xxx", apiKey: "***" });
+    insightsClient("setUserToken", "abc");
+    expect(insightsClient("getUserToken")).toEqual("abc");
+
+    insightsClient("setUserToken", "def");
+    insightsClient("getUserToken", null, (err, token) => {
+      expect(token).toEqual("def");
+      done();
+    });
+  });
+});
--- a/lib/_cookieUtils.ts
+++ b/lib/_cookieUtils.ts
@@ -56,6 +56,7 @@ export function getUserToken(
 ): string {
   if (isFunction(callback)) {
     callback(null, this._userToken);
+    return;
   }
   return this._userToken;
 }
--- a/lib/_getFunctionalInterface.ts
+++ b/lib/_getFunctionalInterface.ts
@@ -4,7 +4,7 @@ import AlgoliaAnalytics from "./insights";
 export function getFunctionalInterface(instance: AlgoliaAnalytics) {
   return (functionName: string, ...functionArguments: any[]) => {
     if (functionName && isFunction((instance as any)[functionName])) {
-      instance[functionName](...functionArguments);
+      return instance[functionName](...functionArguments);
     }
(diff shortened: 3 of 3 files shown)


lib/__tests__/node.test.ts
```diff
@@ -0,0 +1,15 @@
+import insightsClient from "../node";
```
francoischalifour, line 1: Why is this only tested for Node?

lib/__tests__/node.test.ts
```diff
@@ -0,0 +1,15 @@
...
+
+describe("node", () => {
+  it("correctly return userToken", done => {
```
francoischalifour, line 4: We can write two different test cases for the two APIs to maintain them more easily.

lib/__tests__/node.test.ts
```diff
@@ -0,0 +1,15 @@
...
+    insightsClient("init", { appId: "xxx", apiKey: "***" });
+    insightsClient("setUserToken", "abc");
+    expect(insightsClient("getUserToken")).toEqual("abc");
+
+    insightsClient("setUserToken", "def");
+    insightsClient("getUserToken", null, (err, token) => {
```
francoischalifour, line 10: We don't test that the callback form doesn't return anything.

eunjae-lee:
Okay, I thought it was going to be a simple bug fix but I've realized that this cannot be fixed without any breaking change. We do not need this right now. So I'll close this for now.


1. In Node.js env, this PR actually fixes the problem.

2. But in browser env, this doesn't.

```js
var ALGOLIA_INSIGHTS_SRC = "https://cdn.jsdelivr.net/npm/search-insights@1.4.0";

!(function (e, a, t, n, s, i, c) {
  (e.AlgoliaAnalyticsObject = s),
    (e[s] =
      e[s] ||
      function () {
        (e[s].queue = e[s].queue || []).push(arguments);
      }),
    (i = a.createElement(t)),
    (c = a.getElementsByTagName(t)[0]),
    (i.async = 1),
    (i.src = n),
    c.parentNode.insertBefore(i, c);
})(window, document, "script", ALGOLIA_INSIGHTS_SRC, "aa");
```

Users use the snippet above. 

```js
(e[s] =
      e[s] ||
      function () {
        (e[s].queue = e[s].queue || []).push(arguments);
      })
```

can be translated to

```js
(window.aa =
      window.aa ||
      function () {
        (window.aa.queue = window.aa.queue || []).push(arguments);
      })
```

<details>
 <summary>From the very beginning of `search-insights`'s history, `window.aa()` cannot directly return any value to users. It's intended.</summary>
Right after a user puts the snippet into their website, `ALGOLIA_INSIGHTS_SRC` is actually not loaded yet, and will be loaded asynchronously. In the meantime, we prepare `window.aa` but it's just a simple method adding payloads to a queue. And when `ALGOLIA_INSIGHTS_SRC` is loaded, it will consume the queue. That's why `window.aa()` isn't designed to return something directly.
</details>

PR #212, opened 2020-07-20
feat: prevent mutiple initializations

eunjae-lee:
This PR prevents multiple initializations.

### DIFF
--- a/lib/init.ts
+++ b/lib/init.ts
@@ -53,6 +53,11 @@ export function init(options: InitParams) {
       `optional cookieDuration is incorrect, expected an integer.`
     );
   }
+  if (this._initialized === true) {
+    throw new Error(
+      `Already initialized. Do not call 'init' method more than once.`
+    );
+  }
 
   this._apiKey = options.apiKey;
   this._appId = options.appId;
@@ -75,4 +80,6 @@ export function init(options: InitParams) {
   if (supportsCookies()) {
     this.setUserToken(this.ANONYMOUS_USER_TOKEN);
   }
+
+  this._initialized = true;
 }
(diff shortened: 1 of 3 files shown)


lib/init.ts
```diff
@@ -53,6 +53,11 @@ export function init(options: InitParams) {
       `optional cookieDuration is incorrect, expected an integer.`
     );
   }
+  if (this._initialized === true) {
+    throw new Error(
+      `Already initialized. Do not call 'init' method more than once.`
```
Haroenv, line 58: this is a breaking change, we should probably ignore it or update the parameters instead

francoischalifour:
Can you tell us in the PR description why we need this change?

AFAIK there's no other way to change credentials than to init another time.

eunjae-lee:
> Can you tell us in the PR description why we need this change?
> 
> AFAIK there's no other way to change credentials than to init another time.

I've come up with this, when I was thinking of `insights` middleware.

For now, user has to initialize insightsClient by themselves, but once the middleware is introduced, when it's used, it will do almost everything under the hood including the initialization.

When a user moves onto the middleware but forget to remove the `init` function call, it actually won't cause much of trouble, but what would be a good way to alert/prevent it? Should we just print a warning message?

francoischalifour:
Although it's a private API, we store `_hasCredentials` on the Search Insights instance, which becomes `true` when it's initialized.

We can add a warning in the InstantSearch middleware if `_hasCredentials` is `true` to say that it's not needed to initialize the client yourself.

If we start relying on this private API, we might want to make it public.

eunjae-lee:
> Although it's a private API, we store `_hasCredentials` on the Search Insights instance, which becomes `true` when it's initialized.
> 
> We can add a warning in the InstantSearch middleware if `_hasCredentials` is `true` to say that it's not needed to initialize the client yourself.
> 
> If we start relying on this private API, we might want to make it public.

It makes sense to warn in the InstantSearch side. I'll close this, and use `_hasCredentials`.

PR #216, opened 2020-07-23
feat(get): add _get function

eunjae-lee:
This PR adds a method `_get` for internal usage.

Currently we cannot get the instance variables because the main interface is a function (`aa()`).
The initial reason for this functionality is because I need access to `_hasCredentials`.

The reason why I didn't make this function return a value, but make it accept a callback is,
it won't work in umd build.

```js
!function(e,a,t,n,s,i,c){e.AlgoliaAnalyticsObject=s,e[s]=e[s]||function(){
(e[s].queue=e[s].queue||[]).push(arguments)},i=a.createElement(t),c=a.getElementsByTagName(t)[0],
i.async=1,i.src=n,c.parentNode.insertBefore(i,c)
}(window,document,"script",ALGOLIA_INSIGHTS_SRC,"aa");
```

According to the snippet,

`window.aa` is basically a function adding commands to the queue.

```js
function () {
  (e[s].queue = e[s].queue || []).push(arguments);
}
```

In order to have the same signature across the browser and node env, I've decided to pass the callback function.

### DIFF
--- a/lib/__tests__/get.test.ts
+++ b/lib/__tests__/get.test.ts
@@ -1,0 +1,18 @@
+import AlgoliaAnalytics from "../insights";
+
+describe("get", () => {
+  let analyticsInstance;
+  beforeEach(() => {
+    analyticsInstance = new AlgoliaAnalytics({ requestFn: () => {} });
+  });
+
+  it("should get values from instance after init", () => {
+    const callback = jest.fn();
+    analyticsInstance.init({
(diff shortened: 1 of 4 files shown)


francoischalifour:
What about having actual public APIs? (`getAppId`, `getApiKey`, etc.). Having a generic method that returns private data is not very safe because we're not in control of what users can retrieve.

PR #217, opened 2020-07-23
fix(userToken): let user clear the callback for userToken

eunjae-lee:
This PR updates `onUserTokenChange` so that when it receives `undefined`, then it can clear the previous callback and won't throw any exceptions.

Before this, the following code threw an exception:

```js
analyticsInstance.onUserTokenChange(undefined, { immediate: true });
```

### DIFF
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -189,5 +189,29 @@ describe("init", () => {
         expect(callback).toHaveBeenCalledTimes(1);
       });
     });
+
+    describe("nullish or invalid callback", () => {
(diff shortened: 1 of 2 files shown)


eunjae-lee:
Yeah, on the js environment where type is not checked, we'd better have this to prevent the exceptions :)

francoischalifour:
That's weird to fail silently when passed a string.

We could have thrown here if `callback && !isFunction(callback)`.

https://github.com/algolia/search-insights.js/blob/1e5bd42f48530106f62855bba22a0a4b612bce8d/lib/_cookieUtils.ts#L70

PR #224, opened 2020-08-28
fix(bundle): fix mixed exports issue

### LINKED ISSUE #222: README instructions are wrong?
Hey guys, according to your README, it says:

```
const aa = require('search-insights')

aa('init', { ... })
```

However, aa is an object. Am I going crazy and missed something? I've scoured the documentation multiple times.

eunjae-lee:
## Summary

fixes #222

This PR fixes the mixed exports issue.
Since 7880205, `search-insights` started to have both default export and named exports.
There is [an issue](https://github.com/rollup/rollup/issues/1961) related to this in rollup.

I ended up modifying the bundle output.

from
```js
exports.createInsightsClient=createInsightsClient;
exports.default=node
```
to
```js
module.exports=node;
module.exports.createInsightsClient=createInsightsClient;
```

I couldn't find any elegant way to fix this other than this.

## Result

All of the following works fine.

```js
const aa = require("search-insights");

const { createInsightsClient } = require("search-insights");

import aa, { createInsightsClient } from "search-insights";
```


Haroenv:
This seems fine. I guess babel is no option, since it's rollup creating the exports and not babel?

eunjae-lee:
> This seems fine. I guess babel is no option, since it's rollup creating the exports and not babel?
> 
> Would we be able to create an entry point which is already cjs to conform to the shape from the start?

Didn't think of that!
WDYT https://github.com/algolia/search-insights.js/pull/224/commits/304e924439500867cafd4c61dd8c62ee23c837ab ?

Haroenv (approved):

PR #225, opened 2020-09-01
chore(ci): move to CircleCI and update node version to 12

eunjae-lee:
This PR moves from TravisCI to CircleCI and updates the node version to 12.14.1.

(I tried to update the version of ship.js and because node v8 is used in this repo, I couldn't.)

I compared the built files with the ones from `master` branch and they're the same.


package.json
```diff
@@ -8,9 +8,6 @@
     "http": "./empty-module.cjs.js",
     "https": "./empty-module.cjs.js"
   },
-  "engines": {
-    "node": ">=8.16.0"
```
eunjae-lee, line 12: I don't think at this time people are using something lower than 8.16.0.
So I got rid of them. Any concern?

Haroenv (dismissed):

package.json
Haroenv: Seems fine I guess

yannickcr: With this requirement it was not possible to use it with something lower than 8.16.0 anyway.

With the change in Ship.js you'll need node 12 to contribute but users still only need at least node 8.16.0 to use the package, right ?

If that is the case it is ok (not a breaking change) but you should also update `.travis.yml` to make the build on node 12 but test it on node 8 (is that possible on Travis ?)

francoischalifour: Yes, you can specify multiple versions like so:

https://github.com/algolia/create-instantsearch-app/blob/d9f14869ac80e1d4b1cfa87b11d45d903c3e56d6/.travis.yml#L2-L5

eunjae-lee: But does this mean Travis is going to iterate and run tests per version?
For tests, it's okay, but we need to run `yarn shipjs trigger` on Travis, too.
According to the syntax, I don't think it's possible to run `yarn shipjs trigger` only on node v12, while other tests run on multiple versions.

Any idea?

francoischalifour: You could still conditionally run a Bash command in the script, but that'd suck to put logic in there to trigger a release.

eunjae-lee: Agree. If we move it over to CircleCI, then we can smoothly implement it because in CircleCI we can specify node version per job and we can run multiple jobs.

yannickcr: You can use nvm on Travis to only run some specific commands on node v12: `nvm exec 12 yarn shipjs trigger`

eunjae-lee: @yannickcr ahhh that's a good idea. I like it. I'll update this branch to try it out!

.travis.yml
```diff
@@ -17,6 +17,9 @@ env:
 before_install:
 - curl -o- -L https://yarnpkg.com/install.sh | bash
 - export PATH="$HOME/.yarn/bin:$PATH"
+- curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+- export NVM_DIR="$([ -z "${XDG_CONFIG_HOME-}" ] && printf %s "${HOME}/.nvm" || printf %s "${XDG_CONFIG_HOME}/nvm")"
+- [ -s "$NVM_DIR/nvm.sh" ] && \. "$NVM_DIR/nvm.sh"
```
Haroenv, line 22: I think you might not need to curl nvm: https://discuss.circleci.com/t/nvm-does-not-change-node-version-on-machine/28973/14

eunjae-lee: The link you posted is circleci and in this repo we're using travis. So I've gotta install it by myself :)

Haroenv: oops, I forgot this repo uses travis

Haroenv (dismissed):

package.json
eunjae-lee: It's not going to work. With 8.16.0 in .travis.yml, it cannot even install dependencies.
For tests, it only needs v8 but for release it needs v12. And both needs to install devDependencies. It's not something that cannot differentiate with yarn install --production.
Later I'll find some time to move it over to CircleCI rather than struggling with Travis any longer.

.circleci/config.yml
```diff
@@ -0,0 +1,101 @@
...
+
+  "unit tests":
+    <<: *node_v8
+    steps:
+      - checkout
+      - run: *install_yarn_version
+      - restore_cache: *restore_yarn_cache
+      - run:
+          name: Remove Ship.js and install dependencies
+          command: yarn remove shipjs && yarn install
```
eunjae-lee, line 72: If I don't remove shipjs, yarn install fails because the node is v8.

Haroenv (approved):

PR #233, opened 2020-10-28
fix: do not store anonymous token in cookie when user has opted out

eunjae-lee:
## Summary

This PR follows `userHasOptedOut` a step further.
Even with `userHasOptedOut: true`, search-insights created anonymous user token and stored it in the cookie, which is totally not used at all. (I'm not talking about the possibility of that customers could've read that cookie directly for some reason)

In my opinion, this shouldn't be treated as a breaking change. What do you think?

---

ps. I put the people I recently talked about this with as reviewers.

### DIFF
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -1,10 +1,12 @@
 import AlgoliaAnalytics from "../insights";
 import * as utils from "../utils";
+import { getCookie } from "../_cookieUtils";
 
 describe("init", () => {
   let analyticsInstance;
   beforeEach(() => {
     analyticsInstance = new AlgoliaAnalytics({ requestFn: () => {} });
+    document.cookie = `_ALGOLIA=;${new Date().toUTCString()};path=/`;
   });
 
   it("should throw if no parameters is passed", () => {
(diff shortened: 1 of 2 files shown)


Haroenv:
I think it makes sense, although we might need to change opt out to opt in in the next major

Haroenv (approved):

francoischalifour (approved):

tkrugg (approved):

tkrugg:
From my perspective, this does not break any feature because the library does nothing with the content of the cookie when passed `userHasOptedOut: true`. Is the cookie still used by InstantSearch with Insights ?

eunjae-lee:
> From my perspective, this does not break any feature because the library does nothing with the content of the cookie when passed `userHasOptedOut: true`. Is the cookie still used by InstantSearch with Insights ?

Hey @tkrugg , this cookie is still read by InstantSearch with Insights (but not written).

By the way, we need to think a little more about InstantSearch side, because so far we already assumed we always have a token because of anonymous user token. However now there could be really nothing, meaning

hasUserOptedOut: true
+
no explicit token setting

will make a hole. We should prevent events being sent with no token and alert the devs via warning messages and docs. Does it make sense? @Haroenv @francoischalifour @tkrugg

Haroenv:
yes, we should _not send_ when there's no user token, nor anonymous one and error in development for that case

francoischalifour:
@eunjae-lee Makes sense to me.

tkrugg:
> By the way, we need to think a little more about InstantSearch side, because so far we already assumed we always have a token because of anonymous user token. However now there could be really nothing, meaning
> 
> hasUserOptedOut: true
> +
> no explicit token setting
> 
> will make a hole. We should prevent events being sent with no token and alert the devs via warning messages and docs. Does it make sense? @Haroenv @francoischalifour @tkrugg

Hey @eunjae-lee this isn't quite clear to me yet. 
When `hasUserOptedOut: true`, we [already in a situation where events cannot be sent](https://github.com/algolia/search-insights.js/blob/master/lib/_sendEvent.ts#L25).  
Also, an [error is thrown when `userToken` was not supplied](https://github.com/algolia/search-insights.js/blob/master/lib/_sendEvent.ts#L37). So I wouldn't worry about the case you're highlighting.

tkrugg:
FYI documention is being updated to reflect this change here: https://github.com/algolia/doc/pull/5053.

eunjae-lee:
@tkrugg oh you're right 🤦🏻‍♂️ Thank you!

PR #236, opened 2020-12-15
feat(useCookie): skip anonymous userToken if useCookie is false

eunjae-lee:
This PR adds an option `useCookie` to `init` method.
Its default value is `true`, but it needs to be `false` in the next major version to avoid the situation where our customers set cookies in end-users' browsers without knowing it.

##  `hasUserOptedOut` vs `useCookie`

If user denies to be tracked, for example, by clicking "deny" button on a cookie popup, then `userHasOptedOut` should be true, and search-insights won't send any event from the user. So this is per-user setting.

On the other hand, if a website decides not to use anonymous user token in cookie at all, then they can set `useCookie: false`. It means search-insights will not generate/store any anonymous user token in the cookie. So aa will not have any user token initially. `setUserToken` must be called explicitly. Once a `userToken` is set, it will send events.


lib/init.ts
```diff
@@ -61,7 +62,7 @@ export function init(options: InitParams) {
   this._endpointOrigin = options.region
     ? `https://insights.${options.region}.algolia.io`
     : "https://insights.algolia.io";
-
+  this._useCookie = options.useCookie ?? true;
```
samouss, line 65: What's the difference between `options.useCookie` vs `options.userHasOptedOut` ?

lib/_tokenUtils.ts
```diff
@@ -25,29 +25,27 @@ export const getCookie = (name: string): string => {
...
+  if (
+    !foundToken ||
+    foundToken === "" ||
+    foundToken.indexOf("anonymous-") !== 0
+  ) {
+    this._userToken = `anonymous-${createUUID()}`;
+    setCookie(COOKIE_KEY, this._userToken, this._cookieDuration);
   } else {
-    this._userToken = userToken;
+    this._userToken = foundToken;
```
samouss, line 43: Is it expected that we don't call `onUserTokenChangeCallback` anymore with the anonymous token?

eunjae-lee: That's a good point that I missed. I'll fix it 👍🏼

lib/init.ts
eunjae-lee: If user denies to be tracked, for example, by clicking "deny" button on a cookie popup, then `userHasOptedOut` should be `true`, and `search-insights` won't send any event from the user. So this is per-user setting.

On the other hand, if a website decides not to use anonymous user token in cookie at all, then they can set `useCookie: false`. It means `search-insights` will not generate/store any anonymous user token in the cookie. So `aa` will not have any user token initially. `setUserToken` must be called explicitly. Once a userToken is set, it will send events.

Does it make sense to you? Let me know if anything is unclear.

samouss: Yep that make sense!

lib/_tokenUtils.ts
eunjae-lee: https://github.com/algolia/search-insights.js/pull/236/commits/461c9965347af64179de83a78ecaa9cfec2651e1

README.md
```diff
@@ -65,6 +65,7 @@ aa('setUserToken', 'USER_ID');
 | **`apiKey`**      | `string`       | None (required)          | The search API key of your Algolia application |
 | `userHasOptedOut` | `boolean`      | `false`                  | Whether to exclude users from analytics        |
 | `region`          | `'de' \| 'us'` | Automatic                | The DNS server to target                       |
+| `useCookie`       | `boolean`      | `true`                   | Whether to use cookie in browser environment. The anonymous user token will not be set if `false`. |
```
Haroenv, line 68: after this, I think we should release a version with useCookie as true

lib/_tokenUtils.ts
```diff
@@ -25,29 +25,27 @@ export const getCookie = (name: string): string => {
...
-      foundToken === "" ||
-      foundToken.indexOf("anonymous-") !== 0
-    ) {
-      this._userToken = `anonymous-${createUUID()}`;
-      setCookie(COOKIE_KEY, this._userToken, this._cookieDuration);
-    } else {
-      this._userToken = foundToken;
-    }
+export function setAnonymousUserToken(): void {
+  if (!supportsCookies()) {
```
Haroenv, line 29: if `_useCookie` is `false`, supportsCookies should return true, since no error is needed in that case

README.md
eunjae-lee: You mean `false` as a breaking change? It's true by default.

Haroenv: Yes, sorry I meant `false` in the future

eunjae-lee: yeah it should be `false` in the next major version.

Haroenv (approved):
good to go!

PR #238, opened 2020-12-16
fix: use browser requester on browser with bundler

eunjae-lee:
## Summary

This PR adds `browser` key to `package.json` and it will point to a cjs version for browser (not node).

Before this, importing search-insights via bundler imported the cjs version and it used node requester (which is `https` module). Since this PR, the cjs build is split into two: one for browser and another one for node.


lib/node.ts
```diff
@@ -1,10 +1,17 @@
...
 export function createInsightsClient(requestFn: RequestFnType) {
   return getFunctionalInterface(new AlgoliaAnalytics({ requestFn }));
 }
 
-export default createInsightsClient(getRequesterForNode());
+export default createInsightsClient(
+  typeof window !== "undefined"
+    ? getRequesterForBrowser()
+    : getRequesterForNode()
+);
```
Haroenv, line 17: you should add a `browser` key to the package.json instead of doing this kind of conditional

eunjae-lee: If we choose to do that, users will get to use umd build instead of cjs, even when it's loaded via bundlers. What do you think?

Haroenv: you can make a file like this for browser, no?

eunjae-lee: ah, so you're saying like this?

- umd
- cjs for node
- cjs for browser

eunjae-lee:
I wish there's an automated test, but for now, I've done it manually.

- [umd](https://codesandbox.io/s/objective-boyd-vj151?file=/index.html) - nothing really changed for umd, so it works
- [cjs for browser](https://codesandbox.io/s/gifted-wing-8hluv?file=/src/index.js) - from the network panel, you can see the event sent by `sendBeacon`
- [cjs for node](https://codesandbox.io/s/funny-thunder-ub7cf?file=/src/index.js) - works fine

Haroenv:
before this version, would parcel not take module? This set of exports is indeed not easy. How does microbundle / js client do it?

eunjae-lee:
> before this version, would parcel not take module? This set of exports is indeed not easy. How does microbundle / js client do it?

Before this version, parcel took the cjs version from `main` property, but the cjs was only using node requester regardless of the environment. So with parcel, when it sends an event, it used `https` node module, and parcel polyfilled the dependency into the frontend. That's why I'm splitting cjs into node version and browser version.

* [1](https://github.com/developit/unfetch/blob/master/packages/isomorphic-unfetch/package.json#L12-L13) [2](https://github.com/formium/tsdx/blob/57f7dcc9b8e6252b32582b0f1bc37b8dcd44336b/src/templates/basic.ts#L18-L19) [3](https://github.com/algolia/algoliasearch-client-javascript/blob/master/packages/algoliasearch/package.json#L12-L18)

So we'll have

- `jsdelivr`: umd
- `main`: cjs for node
- `browser`: cjs for browser

We do not support ESM at the moment.

Haroenv (approved):
this looks good!

PR #241, opened 2021-01-11
feat(init): accept userToken at init() method

eunjae-lee:
This PR makes it accept `userToken` parameter at `init()` method, which enables users to initialize the insights client with the initial user token.

If it's given, it will not set anonymous user token in the cookie.

related: https://github.com/algolia/instantsearch.js/pull/4608

### DIFF
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -294,5 +294,16 @@ describe("init", () => {
 
       expect(setAnonymousUserToken).not.toHaveBeenCalled();
     });
+
+    it("can set userToken manually afterwards", done => {
+      analyticsInstance.init({ apiKey: "***", appId: "XXX", userToken: "abc" });
+      analyticsInstance.setUserToken("def");
+      expect(setUserToken).toHaveBeenCalledTimes(2);
+      expect(setUserToken).toHaveBeenLastCalledWith("def");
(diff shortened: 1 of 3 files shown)


lib/__tests__/init.test.ts
```diff
@@ -259,4 +259,40 @@ describe("init", () => {
...
+
+    it("shouldn't set anonymous user token to cookie", () => {
+      analyticsInstance.init({
+        apiKey: "***",
+        appId: "XXX",
+        userToken: "abc",
+        useCookie: true
+      });
+      expect(setUserToken).toHaveBeenCalledTimes(1);
+      expect(setUserToken).toHaveBeenCalledWith("abc");
```
Haroenv, line 293: can we add a test with a Manual setUserToken afterwards?

eunjae-lee: https://github.com/algolia/search-insights.js/pull/241/commits/6aa27679a31a1aea86b7f9a5ad8f4d034663d963

Haroenv (approved):
I think this is a sensible way to avoid using cookies

eunjae-lee:
@Haroenv I forgot the README. Can you give it a review? https://github.com/algolia/search-insights.js/pull/241/commits/ea377babc65c9a5a3ce3420660100ba380aff671

Haroenv (approved):

Issue #9, opened 2018-09-10
[types] - expose types


JonasBa:
Right now we expose no interface nor types for the library. We already use them internally, but it would be great to expose them for the public functions.

**List of functions to type:**
- public init: (params: initParams) => void;
- public initSearch: (params: initSearchParams) => void;
- public click: (params?: Partial<ClickReport>) => void;
- public conversion: (params?: Partial<ConversionReport>) => void;
- public search: (params: SearchReport) => void;

mrousavy:
2 Years later still no exposed types. How am I supposed to use this library on a backend server? Just guess the parameters?

Haroenv:
You can use this library with JavaScript, or write a declaration file based on the types in the repository itself @mrousavy. Alternatively we'd also accept a PR gladly which exposes the types directly, we have just prioritised this, since nobody upvoted this issue in two years, hope you understand!

abdelinho24:
Hi, 
Would be nice to prioritize this as it would be helpful for external React projects. 
Thanks.

PR #243, opened 2021-01-15
fix(types): emit and publish type definitions to npm

etienne-martin:
Related to this issue: https://github.com/algolia/search-insights.js/issues/9

- Type definitions are now emitted to the `./dist` folder using `tsc`.
- Types are now included in the tarball when running `npm publish` or `npm pack`.

### DIFF
--- a/empty-module.cjs.d.ts
+++ b/empty-module.cjs.d.ts
@@ -1,0 +1,4 @@
+// shim for node modules in React Native
+// see https://github.com/facebook/react-native/issues/5079
+type empty = {};
+export default empty;
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ tests/screenshots
 tests/production
 tests/data.txt
 .env
+.idea
--- a/index-browser.cjs.d.ts
+++ b/index-browser.cjs.d.ts
@@ -1,0 +1,1 @@
+export * from "./dist/entry-browser-cjs";
--- a/index-node.cjs.d.ts
+++ b/index-node.cjs.d.ts
@@ -1,0 +1,1 @@
+export * from "./dist/entry-node-cjs";
--- a/lib/typings.d.ts
+++ b/lib/typings.d.ts
@@ -1,0 +1,1 @@
+declare module "*/package.json";
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "version": "1.7.0",
   "jsdelivr": "dist/search-insights.min.js",
   "main": "index-node.cjs.js",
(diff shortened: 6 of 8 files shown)


yannickcr (approved):
Thanks for your work @etienne-martin !

I made a small change to separate the `tsconfig` file in two different configurations (one for type checking and one for declaration file generation) so we keep the original behavior for the `yarn type-check` command.

Beside that it seems good to me :slightly_smiling_face:

etienne-martin:
Great, thank you!

eunjae-lee (approved):
@etienne-martin Thank you for your contribution 🎉

empty-module.cjs.d.ts
```diff
@@ -0,0 +1,4 @@
+// shim for node modules in React Native
+// see https://github.com/facebook/react-native/issues/5079
+type empty = {};
+export default empty;
```
Haroenv, line 4: do these not get generated by the tsc, instead, what if we gitignore it?

etienne-martin: They don't because `index-node.cjs.js`, `index-browser.cjs.js` and `empty-module.cjs.js` live outside of the `./lib` folder and are plain JavaScript files.

eunjae-lee: not sure what you mean

etienne-martin: The package.json points to these entry files, and TypeScript will look for the equivalent but with the `.d.ts` extension.

etienne-martin: Ideally I'd get rid of those files sitting at the root of the project and point the package.json to the entry points generated in the `./dist` folder but I didn't want to change everything.

Haroenv: I mean that tsc generates these files at publish time, why do they need to be committed on git?

Haroenv: >  and are plain JavaScript files.

I see now, makes sense! in the future they should be moved to ts

eunjae-lee: Got it. I will work on it next time.

etienne-martin: Another thing is, I'd use the `ts-expect` package against the build output to test that the types are emitted properly. That would ensure that the chain `package.json --> enty-*.js --> entry-*.d.ts` is working properly. 

Because right now, there are unit test for the source code, but there's nothing that checks that the build output is in good health.

eunjae-lee: Wasn't aware of that. Thanks for the information. I will check it out :)

PR #247, opened 2021-03-04
Explicitly export default

Haizzz:
**Summary**: `index-node.cjs.d.ts` does not have a default export, making the [usage in the docs](https://github.com/microsoft/TypeScript/issues/2726) throw type error import has no call signature. Reexport the default
**Problem**: file `index-node.cjs.d.ts` simply `export *` from file `./dist/entry-node-cjs`. `entry-node-cjs.d.ts` have a default export but `export *` does not let the default export flow through as described by [TypeScript issue 2726](https://github.com/microsoft/TypeScript/issues/2726)
```javascript
window.aa('init', {})
          ^^^^^^^^^^^
// this gives a type error as the library doesn't have a default export and therefore no callable signature
// however the code still runs 
```
**Solution**: explicitly reexport the default

### DIFF
--- a/index-node.cjs.d.ts
+++ b/index-node.cjs.d.ts
@@ -1,1 +1,2 @@
 export * from "./dist/entry-node-cjs";
+export { default } from "./dist/entry-node-cjs";


eunjae-lee:
Hello @Haizzz,
Thanks for creating the pull request!

We're planning the v2 in a couple of months, which won't be a big change (useCookie will become false by default).
In the upcoming version, I'm planning to fix this mixed export issue (probably exporting only named things in cjs).
I hope you understand.

I'll close this PR for now.

eunjae-lee (approved):
I think we should go for it since we haven't decided if we're going to change the exports in the upcoming version (and I think having a mixed exports wouldn't be too bad, especially in most cases, people will only import the default insightsClient. So it might not make sense to export it as a named one)

PR #248, opened 2021-03-09
chore(test): fix e2e test

eunjae-lee:
## Summary

This PR fixes the e2e test that might fail when it receives the reponse from alolianet.com host.

### DIFF
--- a/lib/__tests__/harvest.test.ts
+++ b/lib/__tests__/harvest.test.ts
@@ -230,7 +230,8 @@ describe("Integration tests", () => {
       page.on("response", async interceptedRequest => {
         const request = interceptedRequest.request();
 
-        if (request.url().includes(".algolia.net")) {
+        const url = request.url();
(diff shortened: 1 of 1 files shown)


lib/__tests__/harvest.test.ts
```diff
@@ -230,7 +230,8 @@ describe("Integration tests", () => {
       page.on("response", async interceptedRequest => {
         const request = interceptedRequest.request();
 
-        if (request.url().includes(".algolia.net")) {
+        const url = request.url();
+        if (url.includes(".algolia.net") || url.includes(".algolianet.com")) {
```
shortcuts, line 234: Not sure if that matters but as `includes` isn't supported in IE, we could use something like
```suggestion
        if ([".algolia.net", ".algolianet.com"].indexOf(url) >= 0) {
```

shortcuts: Nevermind 🤔

shortcuts (approved):

PR #249, opened 2021-03-09
fix(cookie): make useCookie false by default

eunjae-lee:
## Sumary

This PR make `useCookie` `false` by default.

BREAKING CHANGE: It used to be `true`, but to be GDPR compliant, we're changing it to `false` by default.

For your information, this PR will be merged to `next` (not `master`).
Once it's ready to release the next major version, we will create a PR to merge `next` into `master`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,12 @@ Search Insights lets you report click, conversion and view metrics using the [Al
 
 <!-- tocstop -->
 
+## Notice on the cookie usage
+
+You're looking at the documentation of `search-insights` v2, which is the new major version.
+
+v2 introduces a breaking change which is `useCookie` being `false` by default. If it's `false`, `search-insights` doesn't generate anonymous userToken. It means no event will be sent until `setUserToken` is explicitly called.
+
 ## Getting started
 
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -65,7 +65,7 @@ aa('setUserToken', 'USER_ID');
 | **`apiKey`**      | `string`       | None (required)          | The search API key of your Algolia application |
 | `userHasOptedOut` | `boolean`      | `false`                  | Whether to exclude users from analytics        |
 | `region`          | `'de' \| 'us'` | Automatic                | The DNS server to target                       |
-| `useCookie`       | `boolean`      | `true`                   | Whether to use cookie in browser environment. The anonymous user token will not be set if `false`. When `useCookie` is `false` and `setUserToken` is not called yet, sending events will throw errors because there is no user token to attach to the events. |
+| `useCookie`       | `boolean`      | `false`                  | Whether to use cookie in browser environment. The anonymous user token will not be set if `false`. When `useCookie` is `false` and `setUserToken` is not called yet, sending events will throw errors because there is no user token to attach to the events. |
```
Haroenv, line 68: this warrants a bigger notice in the readme explaining how without setting useCookie users will be counted per visit

eunjae-lee: I thought of adding it to the changelog. Do you think we need a notice on the readme?

Anyway, with `useCookie: false`, we do not generate anonymous user token at all. There is no token until explicitly set. So it's not counted per visit.

Haroenv: yes, it should have a clear explanation in its own header IMO

eunjae-lee: https://github.com/algolia/search-insights.js/pull/249/commits/e0e2e73e5b557e884a9f9eed4ba42417bf414cc9 wdty?

Issue #245, opened 2021-01-18
Unable to tell whether conversion event was successfully received by Algolia


tf3:
As far as I can tell, neither `aa('convertedObjectIDsAfterSearch', ... }` nor `aa('convertedObjectIDs', ...)` returns anything or accepts a callback, so there is no way to tell whether the conversion event successfully received by Algolia. Is my understanding correct? If so, is there any workaround?

Haroenv:
the events are meant to be "fire-and-forget". They are using sendBeacon, which means that the network request still happens, even if the user navigates away from the page. Are you seeing problems with the events not sending?

tf3:
Thanks for the quick response, @Haroenv!

For what it's worth, we're using this on the backend. One issue is just we aren't 100% sure that requests are being sent successfully, and it would make debugging a lot easier if the functions returned a response. Another concern we have is that we might be sending large numbers of requests in quick succession, and we want to be sure all succeed, which it seems isn't possible right now.

eunjae-lee:
Hello @tf3 Thanks for the feedback. I feel your pain. It's definitely not easy to check at the code level if it succeeded. We're planning to improve it but in the meantime you could manually check if the events are successfully coming on our dashboard.

![image](https://user-images.githubusercontent.com/499898/104934975-f4b8db80-59aa-11eb-9c41-c5fe4f2b1f70.png)

tf3:
Thanks @eunjae-lee, that's very helpful for debugging, at least!

samouss:
I've opened an issue similar to what you are asking. Basically the client is designed to work in browser but with recent support of Node it lacks feature like: batch, request lifetime control (see https://github.com/algolia/search-insights.js/issues/235).

PR #250, opened 2021-03-12
feat: add optional callback to event sending methods

eunjae-lee:
## Summary

This PR adds an optional callback to event sending methods:

- viewedObjectIDs
- viewedFilters
- clickedObjectIDsAfterSearch
- clickedObjectIDs
- clickedFilters
- convertedObjectIDsAfterSearch
- convertedObjectIDs
- convertedFilters

Choosing a callback instead of returning a Promise is because the umd version works with a queue.

`aa(method)` doesn't call the actual function, but queues the method and later invokes it as soon as the library is loaded. That's why we can't return a Promise and use the callback pattern.

### DIFF
--- a/lib/__tests__/callback.test.ts
+++ b/lib/__tests__/callback.test.ts
@@ -160,8 +160,13 @@ describe("AlgoliaAnalytics - callback", () => {
       insightsClient._endpointOrigin = "https://not_existing_domain!!!.com";
       aa("clickedObjectIDsAfterSearch", payload, (error, result) => {
         expect(result).toBeUndefined();
-        expect(error).toMatchInlineSnapshot(
-          `[Error: getaddrinfo ENOTFOUND not_existing_domain]`
+        expect(error).toEqual(
+          expect.objectContaining({
+            errno: "ENOTFOUND",
+            code: "ENOTFOUND",
+            syscall: "getaddrinfo",
+            hostname: "not_existing_domain"
+          })
         );
         done();
       });
--- a/lib/__tests__/callback.test.ts
+++ b/lib/__tests__/callback.test.ts
@@ -1,0 +1,170 @@
+import AlgoliaAnalytics from "../insights";
+import { getFunctionalInterface } from "../_getFunctionalInterface";
+import {
+  requestWithSendBeacon,
+  requestWithXMLHttpRequest,
+  requestWithNodeHttpModule
+} from "../utils/request";
+require("dotenv").config();
+
+const payload = {
+  eventType: "click",
+  eventName: "hit-clicked",
+  userToken: "user-id-1",
+  index: "instant_search",
+  queryID: "ebdfbbdd3fba888026416515c26d7954",
+  objectIDs: ["3828012"],
+  positions: [2]
+};
+
+describe("AlgoliaAnalytics - callback", () => {
+  describe("sendBeacon", () => {
+    beforeEach(() => {
+      navigator.sendBeacon = jest.fn();
+    });
+
+    it("calls the callback with no result", done => {
+      const insightsClient = new AlgoliaAnalytics({
+        requestFn: requestWithSendBeacon
+      });
+      const aa = getFunctionalInterface(insightsClient);
+      aa("init", {
+        appId: process.env.APP_ID,
+        apiKey: process.env.API_KEY
+      });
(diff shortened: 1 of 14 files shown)


Haroenv:
why not just return promises from the request method? It's already forwarded to the sendEvent and will be more consistent than callbacks

eunjae-lee:
> why not just return promises from the request method? It's already forwarded to the sendEvent and will be more consistent than callbacks

https://codesandbox.io/s/young-breeze-mqdti?file=/src/index.js

This is a simplified example of search-insights. Do you see a way to make it return promises?

Haroenv:
Ah good point, I forgot about the queue, it would still be possible, but you'd need to make the enqueue a promise that gets resolved when the method is actually called

lib/__tests__/callback.test.ts
```diff
@@ -0,0 +1,175 @@
...
+      });
+      const aa = getFunctionalInterface(insightsClient);
+      aa("init", {
+        appId: "wrong-app-id",
+        apiKey: "wrong-api-key"
+      });
+      insightsClient._endpointOrigin = "https://not_existing_domain!!!.com";
+      aa("clickedObjectIDsAfterSearch", payload, (error, result) => {
+        expect(result).toBeUndefined();
+        expect(error).toEqual({ responseText: "", status: 0, statusText: "" });
```
francoischalifour, line 108: Why don't we have any useful error here?

francoischalifour (commented):
Can you explain _why_ we need this now?

eunjae-lee:
> Can you explain _why_ we need this now?

I forgot to tag:

https://github.com/algolia/search-insights.js/issues/245

> For what it's worth, we're using this on the backend. One issue is just we aren't 100% sure that requests are being sent successfully, and it would make debugging a lot easier if the functions returned a response. Another concern we have is that we might be sending large numbers of requests in quick succession, and we want to be sure all succeed, which it seems isn't possible right now.

lib/__tests__/callback.test.ts
eunjae-lee: I'm relaying the error from `XMLHttpRequest` and it passes that information in such case.

eunjae-lee:
another round of review please ? @Haroenv @francoischalifour

lib/utils/request.ts
```diff
@@ -1,18 +1,50 @@
...
+  data: object,
+  callback?: RequestCallback
+) => void;
 
-export const requestWithSendBeacon: RequestFnType = (url, data) => {
+export const requestWithSendBeacon: RequestFnType = (url, data, callback) => {
   const serializedData = JSON.stringify(data);
   navigator.sendBeacon(url, serializedData);
+  if (callback) {
+    callback();
```
tkrugg, line 12: Here the callback is deceiving I think, because it's not actually being called after the requests, and so you cannot document it as 

> this function gets called when the request is done, and exposes the status


I think it might be a better DX to show a warning explaining we're in the browser and using sendBeacon guaranteeing calls are always fired but not being able to check it with a callback, and not call the callback at all.

eunjae-lee: Yeah it's indeed deceiving because we don't get any callback from sendBeacon. However, I think we should still call the callback for sendBeacon too, because what if they do some logic after sending the events? In the browsers that don't support sendBeacon, it will fallback to XMLHttpRequest, so we'd better have a consistent behavior. What do you think?

tkrugg: we're essentially simulating the callback. It'd be ok if the simulation worked like an actual callback, firing after the requests, however it's not the case here. So the behaviour is radically different based on whether sendBeacon or XMLHttpRequest was picked, internally, and it's essentially a race condition hard to detect. This is what makes me uncomfortable with the API. 

I think as a user, I would prefer not having the callback fire at all and have a warning explaining to be why.

Haroenv (changes requested):
I agree here, while it's technically following the goal of the issue, the callback gets called too early for sendBeacon and might also not make too much sense there (the event happens anyway, even if you never get a callback)

eunjae-lee:
Originally this issue was raised in context of Node.js. Now that we see the behaviors are different between sendBeacon and XMLHttpRequest, I'm closing this PR for now. The callback for Node.js will be dealt separately (probably in [this client](https://github.com/algolia/algoliasearch-client-javascript/tree/master/packages/client-analytics)).

PR #252, opened 2021-04-01
chore: export AlgoliaAnalytics

eunjae-lee:
This PR exports `AlgoliaAnalytics` to create insights client for those who want more control.
Before, it was only possible to choose which requester to use. 

This PR enables https://github.com/algolia/instantsearch.js/pull/4712

### DIFF
--- a/lib/entry-browser-cjs.ts
+++ b/lib/entry-browser-cjs.ts
@@ -2,13 +2,15 @@ import AlgoliaAnalytics from "./insights";
 import { getFunctionalInterface } from "./_getFunctionalInterface";
 import { getRequesterForBrowser } from "./utils/getRequesterForBrowser";
 import { getRequesterForNode } from "./utils/getRequesterForNode";
+import { processQueue } from "./_processQueue";
 import { RequestFnType } from "./utils/request";
 
 export {
+  getRequesterForBrowser,
+  getRequesterForNode,
   AlgoliaAnalytics,
   getFunctionalInterface,
-  getRequesterForBrowser,
-  getRequesterForNode
+  processQueue
 };
 
 export function createInsightsClient(requestFn: RequestFnType) {
--- a/lib/entry-browser-cjs.ts
+++ b/lib/entry-browser-cjs.ts
@@ -2,16 +2,9 @@ import AlgoliaAnalytics from "./insights";
 import { getFunctionalInterface } from "./_getFunctionalInterface";
 import { getRequesterForBrowser } from "./utils/getRequesterForBrowser";
 import { getRequesterForNode } from "./utils/getRequesterForNode";
-import { processQueue } from "./_processQueue";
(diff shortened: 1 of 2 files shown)


lib/entry-browser-cjs.ts
```diff
@@ -2,9 +2,16 @@ import AlgoliaAnalytics from "./insights";
 import { getFunctionalInterface } from "./_getFunctionalInterface";
 import { getRequesterForBrowser } from "./utils/getRequesterForBrowser";
 import { getRequesterForNode } from "./utils/getRequesterForNode";
+import { processQueue } from "./_processQueue";
 import { RequestFnType } from "./utils/request";
 
-export { getRequesterForBrowser, getRequesterForNode };
+export {
+  getRequesterForBrowser,
+  getRequesterForNode,
```
Haroenv, line 10: I don't understand why browser exports the node requester, but that seems to already have been the case?

Haroenv (commented):
reasonable, but not sure if it's needed for the https://github.com/algolia/instantsearch.js/pull/4712 PR since createInsightsClient already is exposed

lib/entry-browser-cjs.ts
eunjae-lee: Clear it doesn't need to export it. It was there by mistake. I'll get rid of it in the v2.

eunjae-lee:
> reasonable, but not sure if it's needed for the [algolia/instantsearch.js#4712](https://github.com/algolia/instantsearch.js/pull/4712) PR since createInsightsClient already is exposed

https://github.com/algolia/instantsearch.js/pull/4712/files#diff-02b3ab8202d2644b709760f62e00f44461af131e42f694fec1f6388ef080af03R2-R5
It needs to use `AlgoliaAnalytics, getFunctionalInterface, getRequesterForBrowser, processQueue`.

eunjae-lee:
This PR has been changed from

"Exporting AlgoliaAnalytics, getFunctionalInterface, getRequesterForBrowser, processQueue"

to

"Exporting AlgoliaAnalytics".

tkrugg (approved):

lib/entry-browser-cjs.ts
tkrugg: Is it still tree shaked off by builder?

Haroenv:
What about importing from /insights in the InstantSearch test?

eunjae-lee:
> What about importing from /insights in the InstantSearch test?

good idea! then we can just close this PR.

https://github.com/algolia/instantsearch.js/pull/4712/commits/70cde712e84bc69ddb0cbd879991ed4ac5a28ab6

eunjae-lee:
Closing this PR since we'll directly import the functions from the source in https://github.com/algolia/instantsearch.js/pull/4712/commits/70cde712e84bc69ddb0cbd879991ed4ac5a28ab6

PR #256, opened 2021-04-20
fix(_get): remove

eunjae-lee:
## Summary

This PR removes `_get` function. It was introduced for internal purpose at insights middleware in InstantSearch.js.

It's not used anymore and even if it's needed in the future, we can add a proper method instead of this generic one.

BREAKING CHANGE

### NOTE

This PR will be merged to `next` branch for v2.


lib/__tests__/harvest.test.ts
```diff
@@ -159,18 +159,6 @@ describe("Integration tests", () => {
         );
         expect(userToken).toEqual("user-id-1");
       });
-
-      it("should get _hasCredentials from the instance", async () => {
```
eunjae-lee, line 163: This test is redundant. We do not need `_hasCredentials` anywhere in search-insights or InstantSearch.js. (The test may have been added to see the possibility of avoiding duplicated init calls, but it's taken care in another PR.)

Haroenv (approved):

shortcuts (approved):

PR #260, opened 2021-04-27
feat: add getVersion method

eunjae-lee:
## Summary

This PR adds `getVersion` method. So far, we've had `version` property in the `AlgoliaAnalytics` class but never exposed a getter method.

In InstantSearch.js, we need this to ensure if the insights middleware is being used with the compatible version of search-insights.

### DIFF
--- a/lib/__tests__/harvest.test.ts
+++ b/lib/__tests__/harvest.test.ts
@@ -64,6 +64,13 @@ describe("Library initialisation", () => {
     });
     expect(analyticsInstance._userToken).not.toBeUndefined();
   });
+
+  it("Should return version", done => {
+    analyticsInstance.getVersion(version => {
+      expect(version).toEqual(expect.stringMatching(/\d+\.\d+\.\d+/));
+      done();
+    });
+  });
 });
 
 describe("Integration tests", () => {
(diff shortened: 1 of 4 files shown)


lib/__tests__/harvest.test.ts
```diff
@@ -64,6 +64,13 @@ describe("Library initialisation", () => {
     });
     expect(analyticsInstance._userToken).not.toBeUndefined();
   });
+
+  it("Should return version", done => {
```
tkrugg, line 68:
```suggestion
  it("should return version", done => {
```

shortcuts (approved):

francoischalifour:
> In InstantSearch.js, we need this to ensure if the insights middleware is being used with the compatible version of search-insights.

Can't we do this with feature detection instead? It feels weird to broaden the API surface area just for this.

eunjae-lee:
> Can't we do this with feature detection instead? It feels weird to broaden the API surface area just for this.

If the way we use search-insights was an object based instead of function based, it would be possible like this:

```js
if (insights.someNewMethod) {
  insights.someNewMethod();
} else {
  ...
}
```

but it's not the case

```js
aa('someMethod')
```

And especially if it's not a new feature but a fix, then feature detection won't work.

https://github.com/algolia/search-insights.js/commit/738e5d9e2a9c416104949ca3509b65e7cb790079

↑ From the point view of InstantSearch.js, it's hard to know if customer is using the version with this fix included or not. Exposing a version seems the most straight way, plus we can warn users to upgrade search-insights to a certain version that way.

Issue #268, opened 2021-05-18
unfriendly type error


eunjae-lee:
The following code is followed by an unfriendly type error:

```js
aa("clickedObjectIDsAfterSearch", {
  userToken: "user-1",
  eventName: "your_event_name",
  index: "your_index_name",
  queryID: "queryID",
  objectIDs: ["objectID1", "objectID2"],
  positions: [17, 19]
});
```

<img width="546" alt="CleanShot 2021-05-18 at 17 02 30@2x" src="https://user-images.githubusercontent.com/499898/118675374-d947cb00-b7fa-11eb-8ba4-b1f6298eda43.png">

reproduction: https://codesandbox.io/s/insights-typing-issue-with-clickedobjectidsaftersearch-oopv8?file=/src/index.ts

It turns out when a method is overloaded, the TypeScript iterates over each overload and checks if the code matches it and shows the result. For [example](https://www.typescriptlang.org/play?#code/C4TwDgpgBAsiBiBXAdgYwIxQLxQBQFsJgALAewBMAuKAIgDMUMaAaKMAQwCd38BnagN4AoKKKjIeEar2CcAlsgDmAbiEBfAJTYAfFABupOeVVDQkWAkYAmbHkIkK1etZZsuPflGFio7cuU4IXk8ZeSVVTR19Q2MhU3BoOCQ0WyTGTAAyC2TUKxNyCFQAGy5oVFJkGSgGNGo0tBMa1FwAcib0FtZvMWIIIqLSahaAd1JOIvIW9Q0gA),

<img width="463" alt="CleanShot 2021-05-18 at 17 00 04@2x" src="https://user-images.githubusercontent.com/499898/118675581-fed4d480-b7fa-11eb-9c6b-08c5d3fbbf51.png">

↑ For that code, we get this ↓

<img width="887" alt="CleanShot 2021-05-18 at 16 59 59@2x" src="https://user-images.githubusercontent.com/499898/118675615-05634c00-b7fb-11eb-8328-d64fd75d7abf.png">

In our case, apparently there are quite a lot of overloads and I think TS is only showing the result for the last overload. Not sure how to improve this yet.

fernandopadang:
i get some issue like this, when i used event 'clickedObjectIDsAfterSearch'.
in documentation we need call event clickedObjectIDsAfterSearch like this
```
aa('clickedObjectIDsAfterSearch', {
  index: 'INDEX_NAME',
  eventName: 'Click item',
[...]

PR #269, opened 2021-05-25
fix(types): provide better error messages for insights client

eunjae-lee:
## Summary

This PR updates the type `InsightsClient`. When the second parameter was wrong, TypeScript gave unhelpful error messages. With this change, now it can infer which method is being used, and thus provide better error messages about the second parameter.

fixes: #268

* sandbox: https://codesandbox.io/s/insights-typing-issue-with-clickedobjectidsaftersearch-forked-wzbon?file=/src/index.ts

---

### example 1

```js
insights("init", { appId, apiKeyy: "aaa" });
```

CleanShot 2021-05-26 at 11 03 51@2x

### example 2

```js
insights("clickedObjectIDsAfterSearch", {
  userToken: "user-1",
  eventName: "your_event_name",
  index: "your_index_name",
  // queryID: "queryID",
  objectIDs: ["objectID1", "objectID2"],
  positions: [17, 19]
});
```

CleanShot 2021-05-26 at 11 04 00@2x

### example 3

```js
insights('unknown_method');
```

CleanShot 2021-05-26 at 11 09 00@2x

Now it infers the method and provide error messages for that specific method.

### DIFF
--- a/lib/_getFunctionalInterface.ts
+++ b/lib/_getFunctionalInterface.ts
@@ -7,6 +7,7 @@ export function getFunctionalInterface(
 ): InsightsClient {
   return (functionName, ...functionArguments) => {
     if (functionName && isFunction((instance as any)[functionName])) {
+      // @ts-ignore
       instance[functionName](...functionArguments);
     } else {
       console.warn(`The method \`${functionName}\` doesn't exist.`);
--- a/lib/click.ts
+++ b/lib/click.ts
@@ -1,6 +1,7 @@
 import { InsightsEvent } from "./_sendEvent";
 
 export interface InsightsSearchClickEvent {
+  eventName: string;
   userToken?: string;
   timestamp?: number;
   index: string;
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -14,84 +14,63 @@ import {
 import { viewedObjectIDs, viewedFilters } from "./view";
 import { getVersion } from "./_getVersion";
 
-export type Init = (method: "init", ...args: Parameters<typeof init>) => void;
+type InsightsMethodMap = {
+  init: Parameters<typeof init>;
+  getVersion: Parameters<typeof getVersion>;
+  addAlgoliaAgent: Parameters<typeof addAlgoliaAgent>;
+  setUserToken: Parameters<typeof setUserToken>;
+  getUserToken: Parameters<typeof getUserToken>;
+  onUserTokenChange: Parameters<typeof onUserTokenChange>;
+  clickedObjectIDsAfterSearch: Parameters<typeof clickedObjectIDsAfterSearch>;
+  clickedObjectIDs: Parameters<typeof clickedObjectIDs>;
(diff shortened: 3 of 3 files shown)


lib/click.ts
```diff
@@ -1,6 +1,7 @@
 import { InsightsEvent } from "./_sendEvent";
 
 export interface InsightsSearchClickEvent {
+  eventName: string;
```
eunjae-lee, line 4: This was missing. It could be a separate PR on its own, but I just added it here.

lib/types.ts
```diff
@@ -14,84 +14,63 @@ import {
 import { viewedObjectIDs, viewedFilters } from "./view";
 import { getVersion } from "./_getVersion";
 
-export type Init = (method: "init", ...args: Parameters<typeof init>) => void;
+export type InsightsMethodMap = {
```
eunjae-lee, line 17: Now we have a map of types of methods.

lib/types.ts
```diff
@@ -14,84 +14,63 @@ import {
...
+type MethodType<MethodName extends keyof InsightsMethodMap> = (
+  method: MethodName,
+  ...args: InsightsMethodMap[MethodName]
 ) => void;
 
-export type AddAlgoliaAgent = (
-  method: "addAlgoliaAgent",
-  ...args: Parameters<typeof addAlgoliaAgent>
-) => void;
+export type Init = MethodType<"init">;
```
eunjae-lee, line 41: And each method is made from the map.

lib/types.ts
```diff
@@ -14,84 +14,63 @@ import {
...
-  ConvertedObjectIDsAfterSearch &
-  ConvertedObjectIDs &
-  ConvertedFilters &
-  ViewedObjectIDs &
-  ViewedFilters;
+export type ViewedObjectIDs = MethodType<"viewedObjectIDs">;
+
+export type ViewedFilters = MethodType<"viewedFilters">;
+
+export type InsightsClient = <MethodName extends keyof InsightsMethodMap>(
```
eunjae-lee, line 73: `InsightsClient` takes the method name (which is one of the keys of the map) as the first parameter, and the second parameter is the exact parameter for the method.

lib/_getFunctionalInterface.ts
```diff
@@ -7,6 +7,7 @@ export function getFunctionalInterface(
 ): InsightsClient {
   return (functionName, ...functionArguments) => {
     if (functionName && isFunction((instance as any)[functionName])) {
+      // @ts-ignore
```
shortcuts, line 10: If this is meant to stay, we could use `@ts-expect-error` and explain why it's needed

lib/click.ts
shortcuts: I think it's fine to have it here since it's still related to typescript

shortcuts (approved):
Looks good to me!

Haroenv (approved):
great change!

Issue #272, opened 2021-06-30
require.js issue with loading from cdn


prayuj:
Hi,
We have loaded search insights the way it has been mentioned in the [documentation](https://www.algolia.com/doc/guides/building-search-ui/going-further/send-insights-events/js/#installing-the-search-insights-library). 
However an intermittent issue occurs `Error: Mismatched anonymous define() module` from require js.
Full description of the error is [here](https://github.com/algolia/search-insights.js/files/6741595/error.txt)
![image](https://user-images.githubusercontent.com/30574783/123977132-5a8aa400-d9dc-11eb-8cc6-58c237656bce.png)
Could please provide a way we can use the library without facing this issue.

Haroenv:
In our examples (https://codesandbox.io/s/github/algolia/doc-code-samples/tree/master/InstantSearch.js/algolia-insights) this doesn't happen, so I think this happens when you might be using requirejs on the page or something similar? Can you reproduce this in a simple page as well, or only with the complete page?

prayuj:
Hi,
This is our feature [link](https://pr-8461-author.pr.eks.penseurope-test.com/ie/).
If you check the browser logs you can see the error. We have implemented it the same way the documentations says to implement, i.e. add the script to head

tkrugg:
Hi @prayuj I can't really tell why, but I can't reproduce your issue. I was able to send a click event.
I've recorded my attempt here, please let me know anything I'm doing wrong.
![2021-07-01 11 06 03](https://user-images.githubusercontent.com/2982512/124097991-69b13680-da5c-11eb-85be-74d7088e6db7.gif)

Haroenv:
I actually do see the error, which mentions that you should use `require` to get the code, so the snippet becomes approximately this (I haven't tested it, since I don't have a RequireJS setup ready):

```js
var ALGOLIA_INSIGHTS_SRC = 'https://cdn.jsdelivr.net/npm/search-insights@1.8.0';

window.AlgoliaAnalyticsObject = s;
window.aa =
  window.aa ||
  function() {
    (window.aa.queue = window.aa.queue || []).push(arguments);
  };

require(ALGOLIA_INSIGHTS_SRC);
```

prayuj:
Hi @Haroenv,
So this is what we have done
```
require(['https://cdn.jsdelivr.net/npm/search-insights@2.0.0'], function (algoliaInsights) {
window.algolia = {};

const algoliaInsightEvents = algoliaInsights.default;
[...]

PR #274, opened 2021-07-09
fix: provide iife build to prevent conflicts with AMD

tkrugg:
This will fix https://github.com/algolia/search-insights.js/issues/272 directly loaded UMD does not play nice with RequireJS.
If you're using the snippet, this issue may appear inconsistely depending on the order in which they're ran. 
Ensure insights is loaded after RequireJS to reproduce it consistenly, like this: 

```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/search-insights@1.7.1/dist/search-insights.min.js"></script>
```
Try it here: https://codesandbox.io/s/elated-lalande-elvc7?file=/index.html:244-449

## Alternative solution
Wrap the UMD build file into (see https://github.com/parcel-bundler/parcel/issues/2451#issuecomment-459590885)

```js
(function(){
   var define = null
   // UMD build content goes here
})()
```
This should be easy enough to do, but the IIFE build is a simpler and smaller solution.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "search-insights",
   "description": "Library for reporting click, conversion and view metrics using the Algolia Insights API",
   "version": "2.0.1",
-  "jsdelivr": "dist/search-insights.min.js",
+  "jsdelivr": "dist/search-insights.iife.min.js",
   "main": "index-node.cjs.js",
   "types": "index-node.cjs.d.ts",
   "browser": "index-browser.cjs.js",
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -54,5 +54,15 @@ export default [
     },
     external: ["http", "https"],
     plugins: createPlugins()
+  },
+  {
+    input: "lib/entry-umd.ts",
+    output: {
+      format: "iife",
+      name: MODULE_NAME,
+      file: `./dist/${LIBRARY_OUTPUT_NAME}.iife.min.js`
+    },
+    external: ["http", "https"],
+    plugins: createPlugins()
   }
 ];


rollup.config.js
```diff
@@ -54,5 +54,15 @@ export default [
     },
     external: ["http", "https"],
     plugins: createPlugins()
+  },
+  {
+    input: "lib/entry-umd.ts",
+    output: {
+      format: "iife",
```
Haroenv, line 61: a umd has to be an iife? this seems wrong to me, that would mean we'd also have to do this for all other libraries

rollup.config.js
```diff
@@ -54,5 +54,15 @@ export default [
...
+      name: MODULE_NAME,
+      file: `./dist/${LIBRARY_OUTPUT_NAME}.iife.min.js`
+    },
+    external: ["http", "https"],
```
Haroenv, line 65: how come this code even needs to be externalised, does the umd build import http and https??

rollup.config.js
tkrugg: I guess you could see UMD a special kind of IIFE that's doing specific things to be compatible with AMD, CommonJS, etc.
The real issue is the fact that we can't load an UMD module via a `<script>` tag if require.js happens to be on that page as well, which is too bad because  that's precisely what the snippet does. 

For most libraries, a UMD build is best because if you do have Require.js on your page, you should use it to fetch and use all your other libraries (InstantSearch.js, search client, etc)

rollup.config.js
tkrugg: When building for node, the http requests are fired by [http/https modules](https://github.com/algolia/search-insights.js/blob/master/lib/utils/request.ts#L29-L30).
I guess this makes sure they're never included in the bundle.

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "search-insights",
   "description": "Library for reporting click, conversion and view metrics using the Algolia Insights API",
   "version": "2.0.1",
-  "jsdelivr": "dist/search-insights.min.js",
+  "jsdelivr": "dist/search-insights.iife.min.js",
```
eunjae-lee, line 5: can we leave it as umd, and let the customer use the specific url for iife?

rollup.config.js
sarahdayan: As @eunjae-lee suggested, we could add the IIFE build instead of replacing the UMD one. Users who need it can import the IIFE build, others won't be impacted.

package.json
tkrugg: what exactly would be the vaiue of keeping an UMD build for the snippet? I'm struggling to find any.

eunjae-lee: To be honest, I have no idea what kind of side-effects we'll possibly see after replacing umd with iife. So my suggestion is, to keep things as-is for all the customers who have been using umd with no problem and add a little iife for those with problem with umd.

Haroenv: agreed, changing from umd to iife could be a bigger change than expected

tkrugg: ok, i changed back jsdelivr to serve umd build in 870526a

Haroenv (approved):

eunjae-lee (approved):
Looks good to me. Are you opening a separate PR for the README update?

sarahdayan (approved):

tkrugg:
Added some documentation of the workaround next to the installation steps for the browser env. 
We talked about adding it to the FAQ but we don't currently have an FAQ.

sarahdayan:
@tkrugg I think this would best fit where you teach how to initialize the client in the docs.

README.md
```diff
@@ -75,6 +75,26 @@ aa('setUserToken', 'USER_ID');
 | `cookieDuration`  | `number`       | `15552000000` (6 months) | The cookie duration in milliseconds            |
 | `userToken`       | `string`       | `undefined` (optional)   | Initial userToken. When given, anonymous userToken will not be set. |
 
+**Note for Require.js users**:
+
+If using [Require.js](https://requirejs.org/), the default build of the library being a UMD build, can conflict. You may experience this issue known as [Mismatched anonymous define() modules](https://requirejs.org/docs/errors.html#mismatch).
```
sarahdayan, line 80:
```suggestion
When using [Require.js](https://requirejs.org/), the default UMD build might conflict and throw with a "Mismatched anonymous define() modules" message. This is a [known Require.js issue](https://requirejs.org/docs/errors.html#mismatch).
```

README.md
```diff
@@ -75,6 +75,26 @@ aa('setUserToken', 'USER_ID');
...
+
+To work around this problem, avoid loading the library via a `define()` call. Instead, change `ALGOLIA_INSIGHTS_SRC` to point to the IIFE build, like following:
```
sarahdayan, line 82: If they use an IIFE, can't they use `define`? I thought the problem was caused by the UMD boilerplate conflicting with Require.js.

tkrugg: I want to avoid recommending it because when loading the lib via AMD, the library won't be able to capture interactions up until the point it's fully loaded.

sarahdayan: Okay, so to clarify:

- Using the UMD build _without_ passing it to `define` crashes because of the anonymous `define` call in there.
- Using the UMD build by passing it to `define` crashes because there's no string ID.
- We want users to use a `<script>` tag and not rely at all on Require.js to load the library to make sure we start stacking events right away.

Is this accurate?

If yes, I would be straight about it in the explanation, right now we justify our solution (using an IIFE without `define`) with the wrong reason (using `define` would work and the initial error would no longer show up).

tkrugg: All true except for this one: 

> Using the UMD build by passing it to define crashes because there's no string ID

Loading the UMD build of insights via `define` [works](https://codesandbox.io/s/busy-flower-fyuqt?file=/index.html:343-710), and recommanded by Require.js. But then you wouldn't be using the snippet.

> I would be straight about it in the explanation, right now we justify our solution (using an IIFE without define) with the wrong reason (using define would work and the initial error would no longer show up).

How about we add something like:
We recommand loading our library via our snippets, to garanty the capture interactions occuring before the libary is fully loaded. 
```suggestion
To work around this problem, avoid loading the library via a `define()` call. Instead, we recommand loading our library via the following snippets, to garanty the capture interactions occuring before the libary has been loaded. 
```

sarahdayan: >Loading the UMD build of insights via define works, and recommended by Require.js. But then you wouldn't be using the snippet.

Okay, so it works but we recommend against it because it doesn't work well for us. So in this case, I'd just drop:

>To work around this problem, avoid loading the library via a `define()` call.

And instead go for:

```suggestion
To work around this problem and ensure you capture all interactions occurring before the library is done loading, change `ALGOLIA_INSIGHTS_SRC` to point to the IIFE build, and load it via a `<script>` tag.
```

eunjae-lee:
To add more context,

### Case 1.
```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>
<script src="https://cdn.jsdelivr.net/npm/search-insights@1.7.1/dist/search-insights.min.js"></script>
```

Screenshot 2021-11-26 at 11 08 49@2x

* error
* no `window.aa` defined

### Case 2.
```html
<script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.6/require.min.js"></script>
...

<script>
var ALGOLIA_INSIGHTS_SRC = "https://cdn.jsdelivr.net/npm/search-insights@2.1.0";

!function(e,a,t,n,s,i,c){e.AlgoliaAnalyticsObject=s,e[s]=e[s]||function(){
(e[s].queue=e[s].queue||[]).push(arguments)},i=a.createElement(t),c=a.getElementsByTagName(t)[0],
i.async=1,i.src=n,c.parentNode.insertBefore(i,c)
}(window,document,"script",ALGOLIA_INSIGHTS_SRC,"aa");
</script>
```
Screenshot 2021-11-26 at 11 09 50@2x

* no error
* `window.aa` is defined
* but the library is not loaded (obviously processQueue is not called yet). Otherwise that wrong usage of `init` would've thrown an error saying appid or apikey is missing.

Issue #319, opened 2021-09-06
Search Insights is not compatible with React Native


jamesxabregas:
When trying to init the search-insights library in a React Native project it dies. After trying unsuccessfully to get this working in my own project, I started a new project based on the Algolia example project for React Native and it also dies as soon as I try to import and init search-insights.

I've tried all the suggestions in the error message about clearing caches and tmp directories. It appears the issue is with loading the code in the file search-insights-browser.cjs.min.js, as I have confirmed that the file is present.

```
error: Error: Unable to resolve module url from /Users/james/dev/algolia-example/node_modules/search-insights/dist/search-insights-browser.cjs.min.js: url could not be found within the project or in these directories:
  node_modules

If you are sure the module exists, try these steps:
 1. Clear watchman watches: watchman watch-del-all
 2. Delete node_modules and run yarn install
 3. Reset Metro's cache: yarn start --reset-cache
 4. Remove the cache: rm -rf /tmp/metro-*
[...]

PR #320, opened 2021-09-07
fix(bundle): remove unnecessary exports from bundles

eunjae-lee:
## Summary

This PR removes

- `getRequesterForNode` from `entry-browser-cjs`
- `getRequesterForBrowser` from `entry-node-cjs`

They seem unnecessary.

It also breaks React Native (#319). It resolves `entry-browser-cjs` but it tries to resolve all the dependencies regarding `getRequesterForNode`, and fails.

This may be seen as a breaking change, but I doubt if there's any case where they resolve `entry-browser-cjs` AND need requester for node. Feel free to create an issue for discussion if you have such needs.


Haroenv (approved):
yep, the entry point doesn't need those, it can still be imported from utils

PR #331, opened 2021-10-15
fix(types): make callback optional at onUserTokenChange

eunjae-lee:
## Summary

This PR update the types to make `callback` optional at `onUserTokenChange`.
The functionality was already there, but the type didn't match.


Haroenv:
Why is the callback optional?

PR #335, opened 2021-11-08
feat: support multiple events

eunjae-lee:
## Summary

This PR enables sending multiple events in a single request.

search-insights has not supported sending multiple events, although [the REST API](https://www.algolia.com/doc/rest-api/insights/#push-events) already does.

- viewedObjectIDs
- viewedFilters
- clickedObjectIDsAfterSearch
- clickedObjectIDs
- clickedFilters
- convertedObjectIDsAfterSearch
- convertedObjectIDs
- convertedFilters

The methods above now can take multiple event payloads as arguments.

Also `sendEvent` method that used to be `protected` is now `public`. Its signature has changed, but it was never documented and an internal method, so I don't consider this a breaking change.

### DIFF
--- a/lib/_sendEvent.ts
+++ b/lib/_sendEvent.ts
@@ -18,8 +18,7 @@ export type InsightsEvent = {
 
 export function makeSendEvent(requestFn: RequestFnType) {
   return function sendEvent(
-    eventType: InsightsEventType,
-    eventData?: InsightsEvent
+    ...eventData: InsightsEvent[]
   ) {
     if (this._userHasOptedOut) {
       return;
@@ -30,24 +29,23 @@ export function makeSendEvent(requestFn: RequestFnType) {
       );
     }
 
-    const event: InsightsEvent = {
-      ...eventData,
-      eventType,
-      userToken: eventData?.userToken ?? this._userToken
-    };
+    const events: InsightsEvent[] = eventData.map(data => ({
+      ...data,
+      userToken: data?.userToken ?? this._userToken
+    }))
 
-    return bulkSendEvent(
+    return sendRequest(
       requestFn,
       this._appId,
       this._apiKey,
       this._uaURIEncoded,
       this._endpointOrigin,
-      [event]
+      events
     );
   };
 }
 
-function bulkSendEvent(
+function sendRequest(
   requestFn: RequestFnType,
   appId: string,
   apiKey: string,
--- a/lib/_sendEvent.ts
+++ b/lib/_sendEvent.ts
@@ -1,20 +1,5 @@
 import { RequestFnType } from "./utils/request";
-
-export type InsightsEventType = "click" | "conversion" | "view";
-export type InsightsEvent = {
-  eventType: InsightsEventType;
-
-  eventName: string;
-  userToken: string;
-  timestamp?: number;
-  index: string;
-
-  queryID?: string;
-  objectIDs?: string[];
-  positions?: number[];
-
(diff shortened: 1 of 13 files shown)


lib/_sendEvent.ts
```diff
@@ -1,25 +1,9 @@
...
-
-  filters?: string[];
-};
+import { InsightsEvent } from './types'
 
 export function makeSendEvent(requestFn: RequestFnType) {
   return function sendEvent(
-    eventType: InsightsEventType,
-    eventData?: InsightsEvent
+    ...eventData: InsightsEvent[]
```
JonathanMontane, line 6: Considering that you have no special logic around the spread operator and that you pass all the events at once to sendRequest, wouldn't it be better a simpler interface if this accepted an array instead?

Additionally, it might make sense to rename it into `sendEvents`.

This would map closer to what we have in other clients too.

JonathanMontane (commented):
I have a few suggestions to improve clarity on this new method. Happy to discuss them with you.

lib/_sendEvent.ts
eunjae-lee: You're right. Other public methods like `aa("clickedObjectIDs", { })` needs to keep its signature, so it makes sense to spread there. However, `sendEvent` was never public, I'm already modifying its signature here, so it should be clearer to rename it to `sendEvents` that accepts an array. 👍🏼

eunjae-lee: https://github.com/algolia/search-insights.js/pull/335/commits/673c206505e9284e6df8d10a3baed5e9b14a4174

lib/insights.ts
```diff
@@ -79,6 +78,8 @@ class AlgoliaAnalytics {
   public getUserToken: typeof getUserToken;
   public onUserTokenChange: typeof onUserTokenChange;
 
+  public sendEvents: ReturnType<typeof makeSendEvents>;
```
JonathanMontane, line 81: Not blocking at all and it can be done in an other PR but it would be great if we could make the typing of the public methods more explicit so that users can see right away what is actually expected instead of a cryptic `ReturnType<typeof somePrivateMethod>`.

JonathanMontane (approved):
LGTM 📫 📦📦📦📦

lib/insights.ts
eunjae-lee: Not sure if I follow you exactly.
Screenshot 2021-11-18 at 12 03 18@2x

End-users are seeing suggestions for each method. Or, are you talking about maintainers?

JonathanMontane: Actually, I mostly meant that for the sendEvents method mostly (because I don't think that vscode is smart enough to resolve ReturnType<typeof makeSendEvents>, but it seems it is not exposed in the codesandbox.

Digging into the issue, I realised that you should also add the method in the InsightsMethodMap type in types.ts if you want it to appear in the types for `aa('sendEvents', [...])`

eunjae-lee: vscode seems smart enough to resolve ReturnType, but yes I missed adding it to `InsightsMethodMap`.
Screenshot 2021-11-18 at 13 25 32@2x

https://github.com/algolia/search-insights.js/pull/335/commits/f0daa73d3429d03d9c738925aee922d9c7a6ad8c

JonathanMontane: Yup that's awesome 🙂

JonathanMontane (approved):

PR #336, opened 2021-11-15
fix: fall back to XHR if sendBeacon returns false

eunjae-lee:
## Summary

`sendBeacon` can return `false` if it fails to add the request to the browser queue.
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon#return_values

In that case, we need to send the event in different way, otherwise we lose it. So we fall back to XHR if `sendBeacon` returns `false`.

### DIFF
--- a/lib/utils/request.ts
+++ b/lib/utils/request.ts
@@ -2,7 +2,9 @@ export type RequestFnType = (url: string, data: object) => void;
 
 export const requestWithSendBeacon: RequestFnType = (url, data) => {
   const serializedData = JSON.stringify(data);
-  navigator.sendBeacon(url, serializedData);
+  if (!navigator.sendBeacon(url, serializedData)) {
+    return requestWithXMLHttpRequest(url, data);
+  }
 };
 
 export const requestWithXMLHttpRequest: RequestFnType = (url, data) => {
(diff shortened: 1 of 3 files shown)


lib/utils/request.ts
```diff
@@ -2,7 +2,9 @@ export type RequestFnType = (url: string, data: object) => void;
 
 export const requestWithSendBeacon: RequestFnType = (url, data) => {
   const serializedData = JSON.stringify(data);
-  navigator.sendBeacon(url, serializedData);
+  if (!navigator.sendBeacon(url, serializedData)) {
+    return requestWithXMLHttpRequest(url, data);
```
JonathanMontane, line 6: Note that this assumes that if the underlying browser supports sendBeacon it supports XMLHttpRequest. Looking at caniuse it seems to be a fairly reasonable assumption but it might be best to check if the browser supports it. Not a blocking issue though (considering that we're talking about very _strange_ and uncommon browsers here).

JonathanMontane (approved):
🪨

lib/utils/request.ts
eunjae-lee: You're right. Even before this PR, we've considered XMLHttpRequest as supported :)

PR #338, opened 2021-11-22
chore: schedule release

eunjae-lee:
## Summary

This PR schedules release every Tuesday 11am (Paris time). It doesn't automatically releases the new version, but just calls `shipjs prepare` to create a release pull-request.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,8 +97,8 @@ jobs:
       - run:
           name: Prepare release
           command: |
-            git config --global user.email "eunjae.lee@algolia.com"
-            git config --global user.name "Eunjae Lee"
+            git config --global user.email "algobot@users.noreply.github.com"
+            git config --global user.name "algobot"
             yarn run release --yes --no-browse
 
 workflows:
@@ -116,7 +116,7 @@ workflows:
   prepare_release_every_tuesday:
     triggers:
       - schedule:
-          cron: '0 9 * * 2'
+          cron: "0 9 * * 2"
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,6 +86,20 @@ jobs:
           name: Trigger a release if the latest commit is a release commit
           command: |
             yarn shipjs trigger
+  "prepare release":
+    <<: *node_v12
+    steps:
+      - checkout
+      - run: *install_yarn_version
+      - restore_cache: *restore_yarn_cache
+      - run: *run_yarn_install
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Prepare release
+          command: |
+            git config --global user.email "eunjae.lee@algolia.com"
+            git config --global user.name "Eunjae Lee"
+            yarn run release --yes --no-browse
 
 workflows:
   version: 2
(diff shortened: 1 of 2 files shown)


ship.config.js
```diff
@@ -2,6 +2,13 @@ const fs = require("fs");
 const path = require("path");
 
 module.exports = {
+  shouldPrepare: ({ releaseType, commitNumbersPerType }) => {
+    const { fix = 0 } = commitNumbersPerType;
+    if (releaseType === 'patch' && fix === 0) {
```
eunjae-lee, line 7: If it's a patch update and there is no fix (which means only `chore` commits), then we do not prepare a release pull-request. This prevents unnecessary pull-requests every week.

.circleci/config.yml
```diff
@@ -99,3 +113,13 @@ workflows:
...
+  prepare_release_every_tuesday:
+    triggers:
+      - schedule:
+          cron: '0 9 * * 2'
+          filters:
+            branches:
+              only:
+                - master
+    jobs:
+      - prepare release
```
JonathanMontane, line 125:
```suggestion
      - prepare release

```

.circleci/config.yml
```diff
@@ -86,6 +86,20 @@ jobs:
...
+    steps:
+      - checkout
+      - run: *install_yarn_version
+      - restore_cache: *restore_yarn_cache
+      - run: *run_yarn_install
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Prepare release
+          command: |
+            git config --global user.email "eunjae.lee@algolia.com"
```
JonathanMontane, line 100: We could put any name here right? If that's the case, I would advocate for no PII if possible, so something like `events-ingestion-team@algolia.com`.

ship.config.js
JonathanMontane: So if I understand correctly, this will prevent Renovate from triggering new release PRs?

JonathanMontane (approved):
LGTM 🚢 
Might be best to have no PIIs but not a blocker for me (as it's your PII).

ship.config.js
eunjae-lee: Even before `shouldPrepare` is called, Ship.js checks if the latest commit message starts with "chore: release v", and thus renovate commits won't trigger new releases.

.circleci/config.yml
eunjae-lee: Yeah actually, the InstantSearch team has a shared email address for this purpose. 

something like `events-ingestion@algolia.com / Algolia`? Not sure if that email address is suitable for this process 🤔

eunjae-lee: @JonathanMontane 

https://github.com/algolia/places/blob/38e8f30fd55cecdc723f096abd8c225e335f4e81/.travis.yml#L4-L5

what about that?

JonathanMontane: probs okay 🙂

eldeloro75017:
> ## Summary
> This PR schedules release every Tuesday 11am (Paris time). It doesn't automatically releases the new version, but just calls `shipjs prepare` to create a release pull-request.

eldeloro75017:
awwwyeah

PR #342, opened 2021-12-07
fix: encodeURIComponent filters

eunjae-lee:
## Summary

This PR encodeURIComponent filter values.

### DIFF
--- a/lib/_sendEvent.ts
+++ b/lib/_sendEvent.ts
@@ -1,5 +1,6 @@
 import { RequestFnType } from "./utils/request";
 import { InsightsEvent } from './types'
+import { isUndefined } from "./utils";
 
 export function makeSendEvents(requestFn: RequestFnType) {
   return function sendEvents(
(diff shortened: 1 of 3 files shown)


marchelbling (approved):
👌

lib/_sendEvent.ts
```diff
@@ -14,10 +15,24 @@ export function makeSendEvents(requestFn: RequestFnType) {
...
+        userToken: data?.userToken ?? this._userToken
+      };
+      if (!isUndefined(filters)) {
+        payload.filters = filters.map(filter => {
+          if (filter.indexOf(":") >= 0) {
+            const parts = filter.split(":");
+            return [parts[0], encodeURIComponent(parts[1])].join(":")
+          } else {
+            filter;
+          }
```
JonathanMontane, line 31: Actually, looking at the behaviour on the insights API side, it seems that we do a decodeURIComponent on the entirety of the filter string, so this could probably be simplified with a `filters.map(encodeURIComponent)`

Code in the go monorepo: https://github.com/algolia/go/blob/master/pkg/searchapi/filter.go#L122

Here's a test:
```
curl -X POST \
https://insights.algolia.io/1/events \
-H 'x-algolia-api-key: ${API_KEY}' --referer 'http://example.com/debug' \
-H 'x-algolia-application-id: ${APP_ID}' -H 'x-algolia-agent: insights-js (2.1.0); insights-gtm (1.2.1)' \
-H "Content-Type: application/json" \
-d '{
    "events": [
        {
          "eventType": "click",
          "eventName": "Product Clicked",
          "index": "products",
          "userToken": "user-123456",
          "timestamp": 1639129812691,
          "filters": ["brand%3ACool%25%20Brand","promotion:20%25"],
          "queryID": "43b15df305339e827f0ac0bdc5ebcaa7"
        }
      ]
    }'

{"status":200,"message":"OK"}
```

JonathanMontane (approved):
Not blocking per se, because I don't think the situation would happen frequently, but this encoding strategy is slightly different from the decoding strategy on the API side.

lib/_sendEvent.ts
marchelbling: I advised that we encoded only the filter value but it's indeed more complex. I overlooked pathological facet names.
Sorry Eunjae, I think that as Jon says, we should just urlencode the all filter.

eunjae-lee: Yeah I think you're right. Last confirmation from @marchelbling

marchelbling: Yes (not sure if you saw my previous message), I overlooked the facet name; let's protect the full filter string.

eunjae-lee: @marchelbling  Oh I didn't see your previous message 👍🏼

JonathanMontane (approved):

PR #343, opened 2021-12-08
docs: fix README regarding onUserTokenChange

eunjae-lee:
## Summary

This PR updates the README regarding `onUserTokenChange`.
The previous one was valid until search-insights@1, but not anymore since v2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -294,7 +294,7 @@ aa('onUserTokenChange', callback, options);
 | `immediate` | `boolean`  | Fire the callback as soon as it's attached     |
 
 ```js
-aa('init', { ... });  // ← This sets an anonymous user token if cookie is available.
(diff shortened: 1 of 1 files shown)


JonathanMontane (approved):
LGTM! It might make sense to also add a note that this should only be done if the user has consented to cookie use for compliance with GDPR. WDYT?

eunjae-lee:
> LGTM! It might make sense to also add a note that this should only be done if the user has consented to cookie use for compliance with GDPR. WDYT?

If a user denied cookie usage, the website has to pass `userHasOptedOut: false` to `aa('init', { ... })` method. `userHasOptedOut` is used at `sendEvents` method. And the rest flow (including user token) stays the same. Let me know if it's not clear :)

PR #344, opened 2021-12-08
refactor: uri-encode user agent right before sending events

eunjae-lee:
## Summary

This PR removes `_uaURIEncoded`, and uri-encodes it right before sending events. Previously it was a string being concatenated as new bits were added. Now it's an array, and gets joined with `;` later on.

### DIFF
--- a/lib/__tests__/_algoliaAgent.test.ts
+++ b/lib/__tests__/_algoliaAgent.test.ts
@@ -4,33 +4,49 @@ jest.mock("../../package.json", () => ({ version: "1.0.1" }));
 
 describe("algoliaAgent", () => {
   let analyticsInstance;
+  let requestFn;
+
   beforeEach(() => {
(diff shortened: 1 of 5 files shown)


JonathanMontane (commented):
I'm not sure I understand what type of bug this fixes 😅. What was the motivator for this change?

eunjae-lee:
> I'm not sure I understand what type of bug this fixes 😅. What was the motivator for this change?

This doesn't fix any bug, but just a refactoring. I didn't like the fact that the source and the transformed variable coexist while they don't have to.

JonathanMontane (approved):

JonathanMontane:
It did provide a minor advantage, which was that the value would be computed only once, which would make the library marginally faster. But I think that's a perf over-optimisation.

eunjae-lee:
> It did provide a minor advantage, which was that the value would be computed only once, which would make the library marginally faster. But I think that's a perf over-optimisation.

yeah good point!

PR #353, opened 2022-02-01
chore: update README

eunjae-lee:
## Summary

This PR updates README to include `sendEvents` method, which was introduced in https://github.com/algolia/search-insights.js/pull/335, but the doc wasn't updated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -421,7 +421,7 @@ aa('sendEvents', [
 | ----------- | ------------------------------- | ---------------------------------------------- |
 | `eventType` | `view`                          | `click` | `conversion` | The name of the index related to the event.    |
 | `eventName` | `string`                        | The name of the event.                         |
-| `userToken` | `string` (optional)             | You can specify userToken per event. |
+| `userToken` | `string` (optional)             | search-insights uses anonymous user token or a token set by `setUserToken` method. You can override it by providing a `userToken` per event object. |
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -419,8 +419,8 @@ aa('sendEvents', [
 
 | Option      | Type                            | Description                                    |
 | ----------- | ------------------------------- | ---------------------------------------------- |
-| `eventType` | `view`                          | `click` | `conversion` | The name of the index related to the event     |
-| `eventName` | `string`                        | The name of the event                          |
+| `eventType` | `view`                          | `click` | `conversion` | The name of the index related to the event.    |
+| `eventName` | `string`                        | The name of the event.                         |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -402,6 +402,27 @@ aa('viewedFilters', {
...
+
+You can send multiple events in a single HTTP request, by using `sendEvents` method.
+
+```js
+aa('sendEvents', [
+  {
+    eventType,
+    eventName,
+    userToken,
+    ...
```
shortcuts, line 415: How come we don't document other parameters?

README.md
```diff
@@ -402,6 +402,27 @@ aa('viewedFilters', {
...
+  }
+]);
+```
+
+| Option      | Type                            | Description                                    |
+| ----------- | ------------------------------- | ---------------------------------------------- |
+| `eventType` | `view`                          | `click` | `conversion` | The name of the index related to the event     |
+| `eventName` | `string`                        | The name of the event                          |
+| `userToken` | `string` (optional)             | You can specify userToken per event. |
```
shortcuts, line 424: this option is required on the doc, is it different on this lib?

README.md
```diff
@@ -402,6 +402,27 @@ aa('viewedFilters', {
...
+    userToken,
+    ...
+  }
+]);
+```
+
+| Option      | Type                            | Description                                    |
+| ----------- | ------------------------------- | ---------------------------------------------- |
+| `eventType` | `view`                          | `click` | `conversion` | The name of the index related to the event     |
+| `eventName` | `string`                        | The name of the event                          |
```
shortcuts, line 423:
```suggestion
| `eventType` | `view`                          | `click` | `conversion` | The name of the index related to the event.    |
| `eventName` | `string`                        | The name of the event.                         |
```

README.md
eunjae-lee: Anonymous user token or the one set by `aa('setUserToken', ...)` will be used by default, but they can override it by providing explicit user token within the event object.

This is the same as,

```js
aa('init', { ..., useCookie: true });

aa('clickObjectIDs', {
  ....
  userToken: 'something-explicit-here'
});
```

Do you think more explanation should be there?

README.md
eunjae-lee: it all depends on what kind of event, like if it's a filter, then `filters` should be there. if it's a click, then objectID should be there, etc.

README.md
shortcuts: Ok make sense! I think it would be nice to have it here indeed

shortcuts (approved):

PR #358, opened 2022-11-16
fix: add default `Content-Type` and `Content-Length` headers to `XMLHttpRequest` instance

Fluf22:
React Native for Android wasn't setting a `Content-Type` header and requests were failing silently since we don't set the `XMLHttpRequest.onerror` callback.
For this second point, I'll create an issue, and we should discuss about what's the right way to handle this by default.

### DIFF
--- a/lib/utils/__tests__/request.test.ts
+++ b/lib/utils/__tests__/request.test.ts
@@ -18,6 +18,7 @@ describe("request", () => {
   const sendBeacon = jest.fn(() => true);
   const open = jest.fn();
   const send = jest.fn();
+  const setRequestHeader = jest.fn();
   const write = jest.fn();
 
   beforeEach(() => {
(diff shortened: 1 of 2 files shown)


shortcuts (approved):
looks great :) GG

jkaho:
@Fluf22 heya 👋 by any chance, do you recall why setting `Content-Length` was necessary here? The linked ticket only mentions `Content-Type`.

We have a customer support ticket that seems to stem from `Content-Length` being set in the XMLHttpRequest function: [CR-6763](https://algolia.atlassian.net/browse/CR-6763). Upon investigation, it seems that this is a [forbidden request header](https://xhr.spec.whatwg.org/#the-setrequestheader()-method). Trying to figure out if this is safe to remove 🙂

Fluf22:
I don't recall it being necessary for the fix, I surely went too far
You can go ahead

PR #361, opened 2022-11-16
fix: handle XMLHttpRequest errors

Fluf22:
🎟️ JIRA Ticket: [DI-612](https://algolia.atlassian.net/browse/DI-612)

In order to avoid failing silently without recourse when an XHR doesn't come through, this PR adds the ability to plug an error callback on the requesting function.

For example, when using `XmlHttpRequest`, this snippet will use `addEventListener('error', onErrorCallback)` to plug the callback to the request:
```js
aa('onError', console.error);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,10 +148,10 @@ const { createInsightsClient } = require("search-insights");
 // OR via the UMD
 const createInsightsClient = window.AlgoliaAnalytics.createInsightsClient;
 
-function requestFn(url, data) {
+function requestFn(url, data, options) {
   const serializedData = JSON.stringify(data);
   const { protocol, host, path } = require("url").parse(url);
-  const options = {
+  const reqOptions = {
     protocol,
     host,
     path,
@@ -164,10 +164,12 @@ function requestFn(url, data) {
 
   const { request: nodeRequest } =
     url.indexOf("https://") === 0 ? require("https") : require("http");
-  const req = nodeRequest(options);
+  const req = nodeRequest(reqOptions);
 
   req.on("error", error => {
     console.error(error);
+    // or use the `onError` callback
+    options?.errorCallback?.(error);
   });
 
   req.write(serializedData);
@@ -422,6 +424,17 @@ aa('sendEvents', [
 | `eventType` | `'view'` \| `'click'` \| `'conversion'` | The name of the index related to the event.    |
 | `eventName` | `string`                        | The name of the event.                         |
 | `userToken` | `string` (optional)             | search-insights uses anonymous user token or a token set by `setUserToken` method. You can override it by providing a `userToken` per event object. |
+
+### Handling request errors
+
+You can set an error callback which will be called whenever an error happens during a request.
+
+```js
+aa('onError', (event) => {
+    console.error(event);
+    throw new Error("Oops!");
+});
+```
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -432,7 +432,6 @@ You can set an error callback which will be called whenever an error happens dur
 ```js
 aa('onError', (event) => {
     console.error(event);
-    throw new Error("Oops!");
 });
 ```
 
--- a/lib/__tests__/_sendEvent.node.test.ts
+++ b/lib/__tests__/_sendEvent.node.test.ts
@@ -28,7 +28,6 @@ describe("_sendEvent in node env", () => {
     const instance = new AlgoliaAnalytics({ requestFn });
     aa = getFunctionalInterface(instance);
     aa("init", credentials);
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -164,10 +164,12 @@ function requestFn(url, data) {
 
   const { request: nodeRequest } =
     url.indexOf("https://") === 0 ? require("https") : require("http");
-  const req = nodeRequest(options);
+  const req = nodeRequest(reqOptions);
 
   req.on("error", error => {
     console.error(error);
```
tkrugg, line 170: Shouldn't we just raise the error object?

lib/__tests__/_sendEvent.node.test.ts
```diff
@@ -28,6 +28,7 @@ describe("_sendEvent in node env", () => {
     const instance = new AlgoliaAnalytics({ requestFn });
     aa = getFunctionalInterface(instance);
     aa("init", credentials);
+    // aa("onError", console.error);
```
tkrugg, line 31: shall we clean this up?

sarahdayan (commented):
Some feedback on the tests, they're all very focused on internals and don't really assess the exposed API, which is what matters most here.

I think it would be worth having a test that resambles this:

```js
test('calls the error callback on error', () => {
  const onError = jest.fn();

  // This `aa` should be mocked so that calls trigger errors
  aa('onError', onError);

  // Call `aa` with a typical method that can fail

  expect(onError).toHaveBeenCalledTimes(1);
  expect(onError).toHaveBeenCalledWith(/* … */);
});
```

This provides more confidence that the API works as intended, is easier to read, and then you can drop all the other tests that don't enforce a public contract. wdyt?

README.md
Fluf22: Yeah, the idea is to add the ability to handle errors without going the breaking change route, if possible 😁

lib/_sendEvent.ts
```diff
@@ -1,6 +1,6 @@
 import { RequestFnType } from "./utils/request";
 import { InsightsEvent } from './types'
-import { isUndefined } from "./utils";
+import {isFunction, isUndefined} from "./utils";
```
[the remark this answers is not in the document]
Fluf22: Yeah, I'm currently working on a PR to update deps as much as possible, and I would like to work a little bit on tooling after that, so we get to the level of our others repos!

Fluf22:
Thank you very much for your input! I updated the tests and cleaned a little as it has been suggested here.
I think your idea is totally valid and makes for a better testing suite. I also made it consistent with the test suite it's in, so it doesn't shift off too much.
Also, I must admit that I'm a little bit hesitating of shipping this update as is: sure, it's breaking change free, but now that I've got another PR ready that updates the deps and Node version so much that it surely must imply a breaking change, I almost want to close this PR and just add a `console.error` in the XHR error handler, like the Node requester already has.

Would you have some advice about this?

lib/utils/request.ts
```diff
@@ -32,10 +51,13 @@ export const requestWithNodeHttpModule: RequestFnType = (url, data) => {
...
     url.indexOf("https://") === 0 ? require("https") : require("http");
-  const req = nodeRequest(options);
+  const req = nodeRequest(reqOpts);
 
   req.on("error", error => {
     console.error(error);
+    if (options && options.errorCallback) {
+      options.errorCallback(error);
+    }
   });
```
Fluf22, line 61: @sarahdayan  This part bothers me: we already have a `console.error` in here, so it's not consistent with the XHR requester.
That's what pushed me to suggest benefiting from the breaking change coming with next PR to just add the same `console.error` in the XHR error handler.
Wdyt?

(you can answer in the PR conversation instead of here if you wish, so the discussion isn't duplicated)

sarahdayan:
>now that I've got another PR ready that updates the deps and Node version so much that it surely must imply a breaking change

Does it though? These updates are likely for dev dependencies, I don't see why this would imply a breaking change. If your tests pass, there's no reason to consider that anything was broken.

Conversely, if updating a dependency does break something, then maybe don't update this dependency? A breaking change because of a dependency update isn't a good reason to bump a major IMO.

>I almost want to close this PR and just add a `console.error` in the XHR error handler, like the Node requester already has.

That's an odd design choice in the first place to me to have a `console.error` here instead of a catchable one. @Haroenv do you have any history on why this was done that way?

Haroenv:
What would you do with the information that you have an error?

benhinchley:
[DI-612](https://algolia.atlassian.net/browse/DI-612) has been marked as won't do, closing this PR. Happy to re-open if we want to revisit it at a later date.

PR #372, opened 2023-01-23
feat: implement client side retries and anonymous id handling

benhinchley:
DAU-199
DAU-200


src/insights.ts
```diff
@@ -32,104 +35,111 @@ function flush(insights: SnippetAlgoliaInsights) {
...
+      throw new Error(
+        "Before calling any other method, you need to initialize the library by calling the 'init' function with appId and apiKey parameters"
+      );
     }
 
     this.beacon.send({
-        timestamp: Date.now(),
-        userToken: this.userToken.getUserToken(),
+      timestamp: Date.now(),
+      userToken: this.userToken.getUserToken() ?? "",
```
benhinchley, line 116: If this method returns undefined I would just return early from this function. An empty `userToken` will result in an API error.

jkaho: Threw an error, but lmk if just returning is better

src/insights.ts
```diff
@@ -9,6 +9,8 @@ type BufferedMethodCall = [string, unknown];
 
 type SnippetAlgoliaInsights = BufferedMethodCall[];
 
+export type InsightsRegion = "de" | "us";
+
```
benhinchley, line 13: Sorry, I think I missed this in the other PR, it makes more sense for this to be defined next to the `InsightsApiBeaconClient` as that is where the type is used. The rest of this code is merely passing the value down to that class.

src/insights.ts
```diff
@@ -85,47 +89,20 @@ class AlgoliaInsights {
     this.beacon.flushAndPurgeEvents();
   }
 
-  private userToken: UserToken;
-  public setUserToken(userToken) {
+  setUserToken(userToken) {
```
wwalser, line 92: Parameter needs a type definition.

src/insights.ts
```diff
@@ -134,54 +111,80 @@ class AlgoliaInsights {
...
+
+      ...event,
+    });
+  }
+
+  private addAlgoliaAgent(agent: string) {
+    this.beacon?.addAlgoliaAgent(agent);
+  }
+
+  [k: string]: any;
```
wwalser, line 187: This is a strange part of the interface. What is it used for?

src/insights.ts
```diff
@@ -55,20 +58,21 @@ class AlgoliaInsights {
     }
 
     const initAction = flushedActions.shift();
+    // eslint-disable-next-line prefer-spread
```
wwalser, line 61: If we have a lint rule which prefers spreads… just use a spread.

```
this.init(...(initAction?.args ?? []))
```

Having written that out, I will admit that it's uglier. It saves the eslint-disable line tho!

src/userToken/userToken.ts
```diff
@@ -1,56 +1,10 @@
...
-    return value;
-  }
-}
+import { ExpiringCookieStore, InMemoryStore } from './stores';
 
 interface Store {
-  read(key: string): string | undefined;
-  write(key: string, value: string): string;
+  read: (key: string) => string | undefined;
+  write: (key: string, value: string) => string;
```
wwalser, line 7: It's not much of an improvement in this case, but you may be interested to see:
```
write: <T extends string>(key: string, value: T) => T
```
This can sometimes matter when you want typescript to maintain identity across a function call.

wwalser (approved):
small nits

src/insights.ts
benhinchley: This is a hangover from my first version of this, can be removed.

src/insights.ts
```diff
@@ -37,13 +39,14 @@ function flush(insights: SnippetAlgoliaInsights) {
...
   private beacon?: InsightsApiBeaconClient;
   private emitter = new EventEmitter();
+  private userToken: UserToken;
 
-  constructor(i: SnippetAlgoliaInsights | AlgoliaInsights) {
+  constructor(i: AlgoliaInsights | SnippetAlgoliaInsights) {
     if ('initialized' in i && i.initialized) {
+      // eslint-disable-next-line no-constructor-return
       return i;
     }
```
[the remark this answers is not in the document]
benhinchley: Ahh okay, I should probably write a comment on why this is here in the code, but will write it here first.

This initialized check is in place due to how the code gets loaded in the browser. The snippet that is installed onto the page by the builder sets up a stubbed version of this object, and then once the script loads we replace that stub with the actual version and flush out buffered method calls.

I had taken it from the [`loadAnalytics` call from Segment's Analtyics.js](https://github.com/segmentio/analytics-next/blob/d2d0459dc5d2f7c6f92c177ef4022fde9b3f5d2e/packages/browser/src/browser/index.ts#L306) code which this is heavily modeled from.

src/insights.ts
benhinchley: It does save an eslint-disable, but also creates a bunch of new type errors.

Though I think we can make a couple of quick changes to this to make it a little more explicit.

src/insights.ts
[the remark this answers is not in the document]
wwalser: Thanks!

PR #376, opened 2023-01-24
chore: add tests to new src section

jkaho:


src/eventEmitter.ts
```diff
@@ -6,7 +6,9 @@ export class EventEmitter {
   emit(type: string, data: any) {
     const handlerFuncs = this.handlers[type];
     if (!handlerFuncs || handlerFuncs.length === 0) {
-      return;
+      throw new Error(
```
benhinchley, line 9: Not having any handlers registered is a valid case, we don't want to start throwing errors if the builder doesn't care about listening to the event.

benhinchley (commented):
Just a minor thing on the zoer handler case, but otherwise this looks great.

src/eventEmitter.ts
jkaho: Right, I see what you mean

benhinchley (approved):

PR #387, opened 2023-02-13

Haroenv:

### DIFF
--- a/examples/instantsearch/instantsearch.html
+++ b/examples/instantsearch/instantsearch.html
@@ -6,17 +6,17 @@
     <script>
       !(function (e, a, t, n, s, i, c) {
         (e.AlgoliaAnalyticsObject = s),
-          (e.aa =
-            e.aa ||
+          (e[s] =
+            e[s] ||
             function () {
-              (e.aa.queue = e.aa.queue || []).push(arguments);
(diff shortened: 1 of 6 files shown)


Haroenv:
I don't think unit test is related, but I also can't replicate on my machine 🤔

examples/instantsearch/instantsearch.html
```diff
@@ -6,17 +6,17 @@
     <script>
       !(function (e, a, t, n, s, i, c) {
         (e.AlgoliaAnalyticsObject = s),
-          (e.aa =
```
benhinchley, line 9: This isn't also incorrect in the current install snippet is it?

benhinchley (approved):
Change looks fine, you will need to cherrypick this commit onto the `v2` branch and make the release from there. `master` is currently in a state of flux between `v2` and `v3`, and not in a releasable state at the moment.

examples/instantsearch/instantsearch.html
Haroenv: yes, this change is unrelated, but the snippet [here](https://github.com/algolia/search-insights.js#1-load-the-library) is correct

aymeric-giraudet (approved):

PR #390, opened 2023-02-16

Haroenv:

FX-2264

### DIFF
--- a/lib/__tests__/_createInsightsClient.test.ts
+++ b/lib/__tests__/_createInsightsClient.test.ts
@@ -1,32 +1,32 @@
-import { createInsightsClient } from '../_createInsightsClient';
+import { createInsightsClient } from "../_createInsightsClient";
 
-describe('createInsightsClient', () => {
+describe("createInsightsClient", () => {
   beforeEach(() => {
     window.AlgoliaAnalyticsObject = undefined;
   });
 
-  it('should return a function', () => {
-    expect(typeof createInsightsClient(() => {})).toBe('function');
+  it("should return a function", () => {
+    expect(typeof createInsightsClient(() => {})).toBe("function");
   });
 
-  it('registers itself to window', () => {
+  it("registers itself to window", () => {
     const aa = createInsightsClient(() => {});
 
-    expect(typeof window.AlgoliaAnalyticsObject).toBe('string');
+    expect(typeof window.AlgoliaAnalyticsObject).toBe("string");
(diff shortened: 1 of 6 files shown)


Haroenv:
same as #387, but for the v2 branch

lib/__tests__/_createInsightsClient.test.ts
```diff
@@ -0,0 +1,32 @@
+import { createInsightsClient } from "../_createInsightsClient";
+
+describe("createInsightsClient", () => {
+  beforeEach(() => {
+    window.AlgoliaAnalyticsObject = undefined;
+  });
+
+  it("should return a function", () => {
+    expect(typeof createInsightsClient(() => {})).toBe("function");
+  });
```
sarahdayan, line 10: Why do you need to test this? Isn't it covered already?

lib/__tests__/_createInsightsClient.test.ts
```diff
@@ -0,0 +1,32 @@
+import { createInsightsClient } from "../_createInsightsClient";
+
+describe("createInsightsClient", () => {
+  beforeEach(() => {
+    window.AlgoliaAnalyticsObject = undefined;
+  });
+
+  it("should return a function", () => {
```
sarahdayan, line 8:
```suggestion
  it("returns a function", () => {
```

lib/__tests__/_createInsightsClient.test.ts
```diff
@@ -0,0 +1,32 @@
...
+  });
+
+  it("registers itself to window", () => {
+    const aa = createInsightsClient(() => {});
+
+    expect(typeof window.AlgoliaAnalyticsObject).toBe("string");
+    expect(window[window.AlgoliaAnalyticsObject!]).toBe(aa);
+  });
+
+  it("doesn't register itself to window if key is already taken", () => {
```
sarahdayan, line 19:
```suggestion
  it("doesn't register itself to window if`AlgoliaAnalyticsObject` is already taken", () => {
```

sarahdayan (approved):

lib/__tests__/_createInsightsClient.test.ts
Haroenv: I'll apply those changes to the `master` branch only

lib/__tests__/_createInsightsClient.test.ts
Haroenv: there was no test for this yet, so I wanted to add at least "hey it works"

PR #391, opened 2023-02-16

Haroenv:

FX-2264


Haroenv:
same as #387 and #390,  but for v1, hopefully this will release correctly 😅

lib/__tests__/_createInsightsClient.test.ts
```diff
@@ -0,0 +1,32 @@
+import { createInsightsClient } from "../_createInsightsClient";
```
sarahdayan, line 1: Same comments as in https://github.com/algolia/search-insights.js/pull/390, feel free to apply or not.

sarahdayan (approved):

PR #398, opened 2023-02-27
feat: allow credentials to be set in event payload

dhayab:
Hi team 👋 ,

Currently, `search-insights.js` is designed to only be instantiated once. This is fine for InstantSearch, but not so much with Autocomplete, which can rely on sources coming from multiple Algolia apps.

This is already an issue, and it can be seen in this Autocomplete example which loads results from 2 Algolia apps: https://codesandbox.io/s/interesting-sun-5pehcn?file=/app.tsx.

This PR allows overriding those Algolia credentials ~by setting `appId` and `apiKey` in the event body. If found, they will replace those defined in the initialisation of the library, only for this event.~ by providing additional parameters to insights methods that result in sending events. While it currently only handles custom headers, in the future it can be extended to do more.

The same change will be done for v1 and v2. By not changing the API and only adding credentials to the event payload, this will help us provide the fix for customers who use the patched versions of `search-insights.js`.


src/insightsAPIBeaconClient.test.ts
```diff
@@ -48,6 +48,44 @@ describe('InsightsApiBeaconClient', () => {
...
+      expect.objectContaining({
+        headers: expect.objectContaining({
+          'X-Algolia-Application-Id': clientOpts.applicationId,
+          'X-Algolia-API-Key': clientOpts.apiKey,
+        }),
+      })
+    );
+  });
+
+  test('it overrides credentials if found in the event payload', () => {
```
sarahdayan, line 66: Can we also test that credentials are removed from the payload, and aren't forwarded in the event body to the Insights API?

sarahdayan (commented):
Eventually it would be ideal to have this tested on the public API. There's no test file for it right now, but maybe it's the opportunity to create it?

src/insightsAPIBeaconClient.test.ts
dhayab: Done in https://github.com/algolia/search-insights.js/pull/398/commits/a58248ff3fcbee5187a4d16254714523088d9d0f

dhayab:
> Eventually it would be ideal to have this tested on the public API. There's no test file for it right now, but maybe it's the opportunity to create it?

@sarahdayan do you have anything specific in mind? Public API methods are just passthroughs to `beacon.send()` which is tested.

benhinchley (commented):
Couple of thoughts, nothing blocking.

When I first read about this requirement (somewhere in slack I think), I was thinking we would add a third optional argument to the `sendEvent` method that allows for bits of the request to be updated (headers, method, whatever) and then exposing that optional argument through all of the different methods.

It separates the areas of concern a bit better, rather than overloading the event object with details.

NOTE: By adding the `appId` and `apiKey` as optional attributes of the vent payload, we are potentially storing those details in localStorage for up to 30 days (`send` first writes to localStorage, before flushing the cache through `emit`)

src/insights.ts
```diff
@@ -5,31 +5,35 @@ import type {
...
 type SnippetAlgoliaInsights = BufferedMethodCall[];
 
-type ObjectIDsAfterSearchEvent = {
-  eventName: string;
-  index: string;
+type BaseEvent<T> = Expand<
+  T & {
+    eventName: string;
+    index: string;
+    appId?: string;
```
sarahdayan, line 20: This should stay consistent with the existing options passed to the constructor.

```suggestion
    applicationId?: string;
```

sarahdayan:
> NOTE: By adding the `appId` and `apiKey` as optional attributes of the vent payload, we are potentially storing those details in localStorage for up to 30 days (`send` first writes to localStorage, before flushing the cache through `emit`)

This is actually a good point that we should test: if events with overriden credentials at the payload level didn't get sent, they will be the next time the user visits the page, when the client gets initialized. Here too, we should verify that the event is sent with the right credentials.

src/insights.ts
dhayab: Ah I didn't notice! I'll update it in this PR and the ones for v1/v2.

dhayab:
@benhinchley 

> When I first read about this requirement (somewhere in slack I think), I was thinking we would add a third optional argument to the `sendEvent` method that allows for bits of the request to be updated (headers, method, whatever) and then exposing that optional argument through all of the different methods.
> 
> It separates the areas of concern a bit better, rather than overloading the event object with request details.

Great suggestion, I'll try switching to this approach.

> NOTE: By adding the `appId` and `apiKey` as optional attributes of the vent payload, we are potentially storing those details in localStorage for up to 30 days (`send` first writes to localStorage, before flushing the cache through `emit`)

We'll still have to store optional arguments with the event payload in order to attribute the event to the right application whenever cache is flushed. Do you think it's cause for concern?

benhinchley:
> We'll still have to store optional arguments with the event payload in order to attribute the event to the right application whenever cache is flushed. Do you think it's cause for concern?

It isn’t a cause of concern, I would consider both the application id and search only api key public information, I was just raising it so there was awareness as the beacon abstracts away the fact that everything is first written to localStorage so can be easily missed (and would be an issue if we were storing sensitive or private information)

src/insightsAPIBeaconClient.ts
```diff
@@ -13,6 +13,9 @@ export type InsightsApiEvent = {
   objectIDs?: string[];
   positions?: number[];
   filters?: string[];
+
+  appId?: string;
```
Haroenv, line 17: same here right?

dhayab: I switched to providing additional headers directly, following Ben's comment.

src/insights.ts
dhayab: I switched to providing custom headers directly on method calls, following Ben's comment.

src/insights.test.ts
```diff
@@ -0,0 +1,112 @@
...
+      // @ts-expect-error
+      insights[methodName](payload, additionalParams);
+
+      expect(fetch).toHaveBeenLastCalledWith(
+        expect.any(String),
+        expect.objectContaining({
+          headers: expect.objectContaining(additionalParams.headers),
+        })
+      );
+    }
```
sarahdayan, line 31: I'm not fond of abstracting too much in test files because it makes them harder to read and see test cases that closely look like the code a user would write.

Right now, I have to squint to understand what's the actual public API.

sarahdayan (approved):

benhinchley (approved):
@dhayab Change looks great, I just made one small commit to thread the optional params all the way through the beacon code, rather than overloading the event type.

PR #399, opened 2023-02-27
feat: pass custom credentials as additional arguments of insights methods (v2)

dhayab:
See https://github.com/algolia/search-insights.js/pull/398

### DIFF
--- a/lib/__tests__/click.test.ts
+++ b/lib/__tests__/click.test.ts
@@ -5,66 +5,109 @@ const credentials = {
   appId: "test"
 };
 
-let analyticsInstance;
+const additionalParameters = {
+  headers: {
+    "X-Algolia-Application-Id": "overrideApp123",
+    "X-Algolia-API-Key": "overrideKey123"
+  }
+};
+
+let analyticsInstance: AlgoliaAnalytics;
+
 beforeEach(() => {
   analyticsInstance = new AlgoliaAnalytics({ requestFn: () => {} });
+  analyticsInstance.init(credentials);
+  (analyticsInstance as any).sendEvents = jest.fn();
 });
 
 describe("clickedObjectIDsAfterSearch", () => {
-  test("Should call sendEvents with proper params", () => {
-    const clickParams = {
-      positions: [1],
-      objectIDs: ["2"],
-      queryID: "testing"
-    };
+  const clickParams = {
+    index: "index1",
+    eventName: "hit clicked",
+    positions: [1],
+    objectIDs: ["2"],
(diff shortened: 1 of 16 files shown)


lib/__tests__/click.test.ts
```diff
@@ -5,66 +5,109 @@ const credentials = {
   appId: "test"
 };
 
-let analyticsInstance;
```
dhayab, line 8: A lot of noise in click / conversion / view tests, but basically it amounts to adding a test case for additional parameters which required a bit of moving stuff around to reuse instances and test payloads.

lib/_sendEvent.ts
```diff
@@ -15,25 +16,27 @@ export function makeSendEvents(requestFn: RequestFnType) {
...
+    const events: InsightsEvent[] = eventData.map((data) => {
+      const { filters, ...rest } = data;
+
       const payload: InsightsEvent = {
         ...rest,
         userToken: data?.userToken ?? this._userToken
       };
       if (!isUndefined(filters)) {
-        payload.filters = filters.map(encodeURIComponent)
+        payload.filters = filters.map(encodeURIComponent);
```
Haroenv, line 27: did prettier here fail on the before already as well?

Haroenv (commented):
👍

lib/types.ts
```diff
@@ -52,17 +52,15 @@ export type GetUserToken = MethodType<"getUserToken">;
...
 
 export type ClickedObjectIDs = MethodType<"clickedObjectIDs">;
 
 export type ClickedFilters = MethodType<"clickedFilters">;
 
-export type ConvertedObjectIDsAfterSearch = MethodType<
-  "convertedObjectIDsAfterSearch"
->;
+export type ConvertedObjectIDsAfterSearch =
+  MethodType<"convertedObjectIDsAfterSearch">;
```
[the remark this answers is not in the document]
dhayab: I'll reinstall the dependencies, might be a mixup due to back and forth switching with v1 / v3.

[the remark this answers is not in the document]
dhayab: Seems like it was not respecting prettier before, this was fixed with prettier when I edited this file.

lib/_sendEvent.ts
dhayab: We don't run prettier in CI, so a failure wouldn't be visible if edited in an environment that does not run prettier on file change / save.

Haroenv (approved):

sarahdayan (approved):
Yes! Can you also update the README?

PR #400, opened 2023-02-27
feat: pass custom credentials as additional arguments of insights methods (v1)

dhayab:
See https://github.com/algolia/search-insights.js/pull/398

### DIFF
--- a/lib/__tests__/click.test.ts
+++ b/lib/__tests__/click.test.ts
@@ -5,25 +5,34 @@ const credentials = {
   appId: "test"
 };
 
-let analyticsInstance;
+const additionalParameters = {
+  headers: {
+    "X-Algolia-Application-Id": "overrideApp123",
+    "X-Algolia-API-Key": "overrideKey123"
+  }
+};
+
+let analyticsInstance: AlgoliaAnalytics;
 beforeEach(() => {
   analyticsInstance = new AlgoliaAnalytics({ requestFn: () => {} });
+  (analyticsInstance as any).sendEvent = jest.fn();
+  analyticsInstance.init(credentials);
 });
 
 describe("clickedObjectIDsAfterSearch", () => {
-  test("Should throw if queryID, objectIDs or positions are not sent", () => {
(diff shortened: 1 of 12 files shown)


lib/__tests__/click.test.ts
```diff
@@ -56,86 +65,117 @@ describe("clickedObjectIDsAfterSearch", () => {
...
+      undefined
+    );
+  });
 
+  test("should call sendEvents with additional params if provided", () => {
+    analyticsInstance.clickedObjectIDsAfterSearch(
+      clickParams,
+      additionalParameters
+    );
     expect((analyticsInstance as any).sendEvent).toHaveBeenCalledWith(
```
Haroenv, line 89: I guess it's consistent, but do you know why they have to be casted?

lib/_sendEvent.ts
```diff
@@ -143,9 +146,21 @@ function bulkSendEvent(
   apiKey: string,
   userAgent: string,
   endpointOrigin: string,
-  events: InsightsEvent[]
+  events: InsightsEvent[],
+  additionalHeaders: InsightsAdditionalEventParams["headers"] = {}
```
Haroenv, line 150: technically query params, but I guess it's equivalent

Haroenv (approved):

lib/__tests__/click.test.ts
dhayab: In v1 `AlgoliaAnalytics.sendEvent()` is a protected method. I could replace this with `// @ts-expect-error` as I did for other assertions in those tests if you prefer.

lib/_sendEvent.ts
dhayab: Yes I wondered but I thought that although sent as query parameters, they are first custom headers.

lib/__tests__/click.test.ts
Haroenv: either way is fine I think

sarahdayan (approved):
Let's also update the README.

PR #401, opened 2023-02-27
feat(v2): allow to `init` again while preserving passed options

sarahdayan:
This introduces a ~~`patch`~~ `partial` option in the `init` method which lets you update only the options that you declare, without resetting the others.

If this works for us, we can backport this in v1, and have it implemented in the upcoming v3.

### DIFF
--- a/lib/init.ts
+++ b/lib/init.ts
@@ -13,6 +13,7 @@ export interface InitParams {
   cookieDuration?: number;
   region?: InsightRegion;
   userToken?: string;
+  patch?: boolean;
 }
 
 /**
@@ -63,15 +64,22 @@ You can visit https://algolia.com/events/debugger instead.`);
 
   this._apiKey = options.apiKey;
   this._appId = options.appId;
-  this._userHasOptedOut = !!options.userHasOptedOut;
-  this._region = options.region;
+
+  this._userHasOptedOut = options.patch
+    ? patchOption(options.userHasOptedOut, this._userHasOptedOut)
+    : !!options.userHasOptedOut;
+  this._region = options.patch
+    ? patchOption(options.region, this._region)
+    : options.region;
   this._endpointOrigin = options.region
     ? `https://insights.${options.region}.algolia.io`
     : "https://insights.algolia.io";
-  this._useCookie = options.useCookie ?? false;
-  this._cookieDuration = options.cookieDuration
-    ? options.cookieDuration
-    : 6 * MONTH;
+  this._useCookie = options.patch
+    ? patchOption(options.useCookie, this._useCookie)
+    : options.useCookie ?? false;
+  this._cookieDuration = options.patch
+    ? patchOption(options.cookieDuration, this._cookieDuration)
+    : options.cookieDuration || 6 * MONTH;
   // Set hasCredentials
   this._hasCredentials = true;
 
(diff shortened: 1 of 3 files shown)


lib/init.ts
```diff
@@ -84,3 +92,7 @@ You can visit https://algolia.com/events/debugger instead.`);
     this.setAnonymousUserToken();
   }
 }
+
+function patchOption<TOption>(option: TOption, fallback: TOption) {
+  return isUndefined(option) ? fallback : option;
```
Haroenv, line 97: Why is the options.patch not in this function too? Might allow for some de duplication of the body

lib/init.ts
```diff
@@ -13,6 +13,7 @@ export interface InitParams {
   cookieDuration?: number;
   region?: InsightRegion;
   userToken?: string;
+  patch?: boolean;
```
dhayab, line 16: A few other ideas that could be more self explanatory: `partial` / `keepSettings`.

dhayab (approved):

lib/init.ts
sarahdayan: Other possibilities:

- `preserveOptions`
- `preservePreviousOptions`
- `keepOptions`
- `keepPreviousOptions`
- `mergeOptions`
- `mergeWithPreviousOptions`

cc @Haroenv @benhinchley

lib/init.ts
sarahdayan: I rewrote it in https://github.com/algolia/search-insights.js/pull/401/commits/87133dbd6d2ee7fc5713b9651251c5cc21440467, wdyt?

lib/init.ts
```diff
@@ -84,3 +91,28 @@ You can visit https://algolia.com/events/debugger instead.`);
     this.setAnonymousUserToken();
   }
 }
+
+type ThisParams = {
+  _userHasOptedOut: InitParams["userHasOptedOut"];
+  _useCookie: InitParams["useCookie"];
+  _cookieDuration: InitParams["cookieDuration"];
+  _region: InitParams["region"];
+};
```
sarahdayan, line 100: No mapped types in this version of TypeScript afaik.

lib/init.ts
```diff
@@ -84,3 +91,28 @@ You can visit https://algolia.com/events/debugger instead.`);
...
+function setOptions(
+  target: ThisParams,
+  { patch, ...options }: InitParams,
+  defaultValues: ThisParams
+) {
+  if (!patch) {
+    Object.assign(target, defaultValues);
+  }
+
+  for (const key in options) {
```
Haroenv, line 111: for in needs a `hasOwnProperty` check. Can we use Object.assign or is that too low browser support?

sarahdayan: `Object.assign` is fine, we have a polyfill loaded.

Is https://github.com/algolia/search-insights.js/pull/401/commits/2a3cd27a282fd426085b642b10706350965dccbd what you had in mind?

lib/init.ts
sarahdayan: I like `partial` too: https://github.com/algolia/search-insights.js/pull/401/commits/84872c252f9ca5dfcb12a0fabd47604825a1c13e

benhinchley (approved):

dhayab (approved):

Haroenv (approved):

PR #415, opened 2023-03-07
feat: allow to `init` without credentials (v2)

FabienMotte:
### Summary

This PR allows to `init` without credentials (`appId` and `apiKey`).

You can now pass credentials directly when sending events, so it's not always necessary to set credentials when calling `init`.

### Result

- `init` can be called without `options`, an empty `options` object or defined `options`.
- `sendEvents` will now throw an error if no credentials coming from `init` or custom ones (coming from `additionalParams`) are provided.
- `partial: true` can be used to update `appId` and/or `apiKey`.

### DIFF
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -1,6 +1,6 @@
 import AlgoliaAnalytics from "../insights";
 import * as utils from "../utils";
-import { getCookie } from "../_tokenUtils";
+import { getCookie, MONTH } from "../_tokenUtils";
 
 describe("init", () => {
   let analyticsInstance: AlgoliaAnalytics;
@@ -60,12 +60,11 @@ describe("init", () => {
   });
   it("should use 6 months cookieDuration by default", () => {
     analyticsInstance.init({ apiKey: "***", appId: "XXX" });
-    const month = 30 * 24 * 60 * 60 * 1000;
-    expect(analyticsInstance._cookieDuration).toBe(6 * month);
+    expect(analyticsInstance._cookieDuration).toBe(6 * MONTH);
   });
   it.each(["not a string", 0.002, NaN])(
     "should throw if cookieDuration passed but is not an integer (eg. %s)",
-    cookieDuration => {
+    (cookieDuration) => {
       expect(() => {
         (analyticsInstance as any).init({
           cookieDuration,
@@ -84,6 +83,15 @@ describe("init", () => {
       cookieDuration: 42
     });
     expect(analyticsInstance._cookieDuration).toBe(42);
+  });
+  it("should set cookie parameters if defined in `init`", () => {
+    expect(analyticsInstance._useCookie).toBe(false);
+    expect(analyticsInstance._cookieDuration).toBe(6 * MONTH);
+
+    analyticsInstance.init({ useCookie: true, cookieDuration: MONTH })
+
+    expect(analyticsInstance._useCookie).toBe(true);
+    expect(analyticsInstance._cookieDuration).toBe(MONTH);
   });
   it("should set _endpointOrigin on instance to https://insights.algolia.io", () => {
     analyticsInstance.init({ apiKey: "***", appId: "XXX" });
@@ -419,7 +427,7 @@ describe("init", () => {
       expect(setAnonymousUserToken).not.toHaveBeenCalled();
     });
 
-    it("can set userToken manually afterwards", done => {
+    it("can set userToken manually afterwards", (done) => {
       analyticsInstance.init({ apiKey: "***", appId: "XXX", userToken: "abc" });
       analyticsInstance.setUserToken("def");
       expect(setUserToken).toHaveBeenCalledTimes(2);
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -64,7 +64,7 @@ describe("init", () => {
   });
   it.each(["not a string", 0.002, NaN])(
     "should throw if cookieDuration passed but is not an integer (eg. %s)",
-    (cookieDuration) => {
+    cookieDuration => {
       expect(() => {
         (analyticsInstance as any).init({
           cookieDuration,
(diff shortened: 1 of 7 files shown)


lib/init.ts
```diff
@@ -65,9 +50,6 @@ export function init(options: InitParams) {
 You can visit https://algolia.com/events/debugger instead.`);
   }
 
-  this._apiKey = options.apiKey;
-  this._appId = options.appId;
-
```
dhayab, line 70: I think you still need to map `options.apiKey` to `_apiKey` and the same for `_appId` in `setOptions()`.

FabienMotte: Aren't they supposed to be assigned by the `reduce` in `setOptions()`?

```js
Object.assign(
  target,
  Object.keys(options).reduce(
    (acc, key) => ({ ...acc, [`_${key}`]: options[key] }),
    {}
  )
);
```

dhayab: Oh you're right, I didn't notice this!

dhayab (approved):

sarahdayan (commented):
One issue is that the customer could technically forget to call `init` because they could now always pass credentials in headers (thus not getting an error thrown) but the other options, such as `this._endpointOrigin`, would not have been set and be `undefined` when making the call.

A simple solution could be an internal property `this._hasBeenInitialized` that we set to `true` in `init` and check before sending events. This kinda replaces `this._hasCredentials` which can now be inlined. wdyt?

lib/__tests__/init.test.ts
```diff
@@ -292,6 +276,23 @@ describe("init", () => {
...
+    expect(analyticsInstance._apiKey).toBe("apiKey1");
+
+    analyticsInstance.init({
+      appId: "appId2",
+      partial: true
+    });
+
+    expect(analyticsInstance._appId).toBe("appId2");
+    expect(analyticsInstance._apiKey).toBe("apiKey1");
+  });
```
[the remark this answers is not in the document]
FabienMotte: I'm testing if the old `apiKey` is kept when using `partial: true`, in the previous test, we redefined both `appId` and `apiKey`.

dhayab (approved):

lib/init.ts
```diff
@@ -23,22 +23,7 @@ export interface InitParams {
  * Binds credentials and settings to class
  * @param options: initParams
  */
-export function init(options: InitParams) {
-  if (!options) {
-    throw new Error(
-      "Init function should be called with an object argument containing your apiKey and appId"
```
Haroenv, line 29: these errors should likely be returned right? or do we want people to send init without credentials still (as a way to set the other parameters)? Maybe they can be only dropped if `partial` is true

lib/_sendEvent.ts
```diff
@@ -10,9 +10,12 @@ export function makeSendEvents(requestFn: RequestFnType) {
     if (this._userHasOptedOut) {
       return;
     }
-    if (!this._hasCredentials) {
+    const hasCustomCredentials =
+      additionalParams?.headers["X-Algolia-Application-Id"] &&
+      additionalParams?.headers["X-Algolia-API-Key"];
```
Haroenv, line 15:
```suggestion
    const hasCustomCredentials =
      (this.appId && this.apiKey) || (
      additionalParams?.headers["X-Algolia-Application-Id"] &&
      additionalParams?.headers["X-Algolia-API-Key"]);
```

something like this would mean we can drop _hasCredentials completely

lib/insights.ts
```diff
@@ -54,9 +54,9 @@ class AlgoliaAnalytics {
   _apiKey: string;
   _appId: string;
   _region: string;
-  _endpointOrigin: string;
+  _endpointOrigin = "https://insights.algolia.io";
   _userToken: string;
-  _userHasOptedOut: boolean;
+  _userHasOptedOut = false;
   _useCookie: boolean;
   _cookieDuration: number;
```
sarahdayan, line 61: Don't we wnat to set defaults for those two as well?

FabienMotte: I think we can leave them as `undefined` by default since `_useCookie` is only used in `init()` and `_cookieDuration` only used in `setAnonymousUserToken()` which is used in `init()`, WDYT?

sarahdayan: My point is, if we say that we no longer need to `init` to use Insights, these defaults need to be defined no? `init` just becomes a way to change those options, but they need to be defaulted somewhere.

lib/init.ts
dhayab: I think it's fine to drop those here as a similar error will be thrown with `sendEvent` when there are no credentials.

lib/insights.ts
FabienMotte: It makes sense, thanks @sarahdayan!

lib/init.ts
```diff
@@ -65,9 +50,6 @@ export function init(options: InitParams) {
 You can visit https://algolia.com/events/debugger instead.`);
   }
 
-  this._apiKey = options.apiKey;
-  this._appId = options.appId;
-
   setOptions(this, options, {
```
sarahdayan, line 53: I guess we can update this not to specify the defaults now, since they're assigned in the constructor, no? `setOptions` should only change what was passed.

FabienMotte: I'm not sure to understand what you mean, should we keep these two lines?

Haroenv: currently, if you call `init` without `partial`, it will set the default value for the keys you didn't specify, if those are already set by something else, we probably want to reset them to the default if partial isn't given to preserve the previous behaviour, right? 

in v3 we could make partial the default and not an option, and then we don't need this anymore though

Haroenv (approved):
I think what's currently in the PR is the correct behaviour

lib/init.ts
sarahdayan: Yes that's correct.

sarahdayan (approved):

Haroenv:
Title of the PR was possibly a bit outdated, as this now also allows sendEvent without init (if the event sends their credentials)

FabienMotte:
> Title of the PR was possibly a bit outdated, as this now also allows sendEvent without init (if the event sends their credentials)

Correct, I guess we can update that in the future changelog?

PR #416, opened 2023-03-07
chore: update prettier and fix linting

FabienMotte:
This PR updates `prettier` dependency (and related dependencies) and fix linting.

Wait for #415 to be merged first.


sarahdayan (approved):
Let's merge the target PR first so this doesn't end up in the diff.

package.json
```diff
@@ -73,7 +73,7 @@
     "lunr": "2.3.6",
     "markdown-toc": "1.2.0",
     "node-sass": "4.13.1",
-    "prettier": "1.18.2",
+    "prettier": "2.8.4",
```
dhayab, line 76: (putting my comment here for the sake of having a threaded discussion)

Since you're updating prettier, could you also wrap the prettier script glob patterns in quotes?

```
"prettier": "prettier 'lib/**/*.ts' 'examples/**/*' --write",
```

I noticed they are wrongly extrapolated when passed through yarn. This will allow you to fix a few additional files as well.

dhayab (approved):

PR #422, opened 2023-03-15
feat: make `init` optional (v1)

FabienMotte:
### Summary

[FX-2282](https://algolia.atlassian.net/browse/FX-2282)

This PR makes `init` optional as you can already pass credentials directly when sending events using `sentEvent`, so it's not always necessary to set credentials when calling `init`.

### Result

- `init` can be called without `options`, an empty `options` object or defined `options`.
- `sendEvent` will now throw an error if no credentials coming from `init` or custom ones (coming from `additionalParams`) are provided.
- `partial: true` can be used to update `appId` and/or `apiKey`.
- update the README to reflect the changes.

[FX-2282]: https://algolia.atlassian.net/browse/FX-2282?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/_initSearch.ts
+++ b/lib/_initSearch.ts
@@ -1,3 +1,5 @@
+import { isUndefined } from "./utils";
+
 export interface InitSearchParams {
   getQueryID: () => string | number;
   hitsContainer: string | string[];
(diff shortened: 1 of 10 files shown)


lib/_initSearch.ts
FabienMotte, line 1: I'm not sure what's the purpose of this file since `getQueryID` is not used elsewhere

dhayab (approved):

aymeric-giraudet (approved):

PR #429, opened 2023-03-20
ci: npm publish to branch-specific latest tag

dhayab:
This PR changes the Ship.js config so that future releases of search-insights@v1 are tagged `latest-1.x` on npm, instead of polluting the `latest` tag which should be left for the actual latest version of the library.

### DIFF
--- a/ship.config.js
+++ b/ship.config.js
@@ -4,6 +4,8 @@ const path = require("path");
 module.exports = {
   buildCommand: () => "yarn build && /bin/bash ./pre-deploy.sh",
   pullRequestTeamReviewers: ["frontend-experiences-web"],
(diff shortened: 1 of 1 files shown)


sarahdayan (commented):
Is the pattern `latest-n.x` conventional?

dhayab:
> Is the pattern `latest-n.x` conventional?

I don't know if it is, although it is consistent with a `latest-0.x` tag we have on npm for this package: https://www.npmjs.com/package/search-insights?activeTab=versions.

In reality I don't think people will ever use it, but at least it won't have an impact on `@latest`.

sarahdayan (approved):
I'm good with consistency :)

PR #435, opened 2023-03-29
connector: make userToken optional and add the ability to remove an existing userToken

JasonBerry:


src/userToken/userToken.ts
```diff
@@ -46,9 +45,27 @@ export class UserToken {
...
+    } else {
+      this.userTokenStore.write(USER_TOKEN_KEY, userToken);
+      this.anonmyousIdStore?.delete(ANONYMOUS_ID_KEY);
+    }
+  }
+
+  removeUserToken() {
+    this.userTokenStore.delete(USER_TOKEN_KEY);
     this.anonmyousIdStore?.delete(ANONYMOUS_ID_KEY);
+    if (!this.opts.anonmyousId?.enabled) {
```
wwalser, line 66: On L52, we check `!this.anonmyousIdStore` before creating an `anonmyousIdStore`. Here though, it seem that we want `this.opts.anonmyousId?.enabled` to be true before deleting the store. I'm not sure that part of the API is clear to me.

wwalser (changes requested):
In the `userToken` class, `this.opts.anonmyousId` is misspelled.

src/userToken/userToken.ts
JasonBerry: This just gets things back to the original state: the customer's website created an instance of UserToken with anonymous disabled so we delete the anonymousIdStore (if it's been created by the customer's website calling `setUserToken()`, which we treat as an explicit "I know I said I don't want to use anonymous tokens but I'm asking for one, so just give it to me")

src/userToken/userToken.ts
```diff
@@ -46,9 +45,27 @@ export class UserToken {
...
-  setUserToken(userToken: string) {
-    this.userTokenStore.write(USER_TOKEN_KEY, userToken);
+  setUserToken(userToken?: string) {
+    if (!userToken) {
+      this.userTokenStore.delete(USER_TOKEN_KEY);
+
+      if (!this.anonmyousIdStore) {
+        this.anonmyousIdStore = new ExpiringCookieStore(
+          this.opts.anonmyousId?.lease
+        );
```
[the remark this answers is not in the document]
JasonBerry: This handles the case where you initialise UserToken with anonymousIdStore enabled then call `getUserToken` without having called `setUserToken` (which you would call if you have an userToken you want to store).

wwalser (approved):

PR #441, opened 2023-04-18
feat(userToken): generate in-memory anonymous user tokens by default

dhayab:
This PR introduces a new initialization parameter: `anonymousUserToken`. This boolean (set to `true` by default) generates an in-memory anonymous user token which is sent with every Insights event.

See [FX-2303](https://algolia.atlassian.net/browse/FX-2303) for more information.

[FX-2303]: https://algolia.atlassian.net/browse/FX-2303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/_sendEvent.ts
+++ b/lib/_sendEvent.ts
@@ -20,6 +20,10 @@ export function makeSendEvents(requestFn: RequestFnType) {
       );
     }
 
+    if (!this._userToken && this._anonymousUserToken) {
+      this.setAnonymousUserToken(true);
+    }
+
     const events: InsightsEvent[] = eventData.map((data) => {
       const { filters, ...rest } = data;
 
--- a/lib/_tokenUtils.ts
+++ b/lib/_tokenUtils.ts
@@ -26,7 +26,12 @@ export const getCookie = (name: string): string => {
   return "";
 };
 
-export function setAnonymousUserToken(): void {
+export function setAnonymousUserToken(inMemory = false): void {
+  if (inMemory) {
+    this.setUserToken(`anon-${createUUID()}`);
+    return;
+  }
+
   if (!supportsCookies()) {
     return;
   }
--- a/lib/_tokenUtils.ts
+++ b/lib/_tokenUtils.ts
@@ -28,13 +28,14 @@ export const getCookie = (name: string): string => {
 
 export function setAnonymousUserToken(inMemory = false): void {
   if (inMemory) {
-    this.setUserToken(`anon-${createUUID()}`);
+    this.setUserToken(`anonymous-${createUUID()}`);
     return;
   }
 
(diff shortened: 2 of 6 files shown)


lib/_tokenUtils.ts
```diff
@@ -26,7 +26,12 @@ export const getCookie = (name: string): string => {
   return "";
 };
 
-export function setAnonymousUserToken(): void {
+export function setAnonymousUserToken(transient = false): void {
+  if (transient) {
+    this.setUserToken(`transient-${createUUID()}`);
```
dhayab, line 31: I went with the `transient` prefix as a way to distinguish an in-memory user token from a cookie-based one.

lib/_tokenUtils.ts
```diff
@@ -26,7 +26,12 @@ export const getCookie = (name: string): string => {
   return "";
 };
 
-export function setAnonymousUserToken(): void {
+export function setAnonymousUserToken(inMemory = false): void {
+  if (inMemory) {
+    this.setUserToken(`anon-${createUUID()}`);
```
dhayab, line 31: Used the `anon-` prefix here to match with the implementation in v3.

Haroenv: even though we already use the anonymous- prefix for cookie-stored tokens? Should we be able to distinguish between the two? if so maybe the prefix should be different?

lib/_sendEvent.ts
```diff
@@ -20,6 +20,10 @@ export function makeSendEvents(requestFn: RequestFnType) {
       );
     }
 
+    if (!this._userToken && this._anonymousUserToken) {
+      this.setAnonymousUserToken(true);
+    }
```
Haroenv, line 25: interesting time to do this check, but it does make sense that you want to do it as late as possible.

Haroenv (approved):
I'm on the fence about the `anon-` prefix, but it's not blocking

lib/_tokenUtils.ts
dhayab: I think from an analytics perspective it doesn't matter what prefix is used, as long as the token itself stays the same during is lifecycle. I just went with what's in v3, but I also initially wanted to name it `transient-xxx` so I'm not really locking in on this.

An advantage of having a different prefix than the cookie-based token is that it makes it easier to identify the type of user token in logs.

aymeric-giraudet (approved):

Haroenv (approved):

francoischalifour:
It's really confusing that this new `anonymousUserToken` param is a boolean while `userToken` is a string.

PR #483, opened 2023-07-04
feat(request): enable requests to be awaited upon

JasonBerry:
This will enable being able to `await` on events being sent.

### DIFF
--- a/lib/__tests__/_getFunctionalInterface.test.ts
+++ b/lib/__tests__/_getFunctionalInterface.test.ts
@@ -5,7 +5,9 @@ describe("_getFunctionalInterface", () => {
   let aa;
 
   beforeEach(() => {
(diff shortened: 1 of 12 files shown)


wwalser:
My initial intuition was that we should provide the conventional Node interface and offer the ability to pass a callback but that would break the strict polymorphic type cohesion between node and browser versions of the library.

This looks correct. Have you manually tested that the code form the ticket executes as expected now? https://algolia.atlassian.net/browse/CR-3696?focusedCommentId=316055

wwalser (approved):
LGTM

Haroenv (approved):

JasonBerry:
> Have you manually tested that the code form the ticket executes as expected now?
I eyeballed the code from the ticket and it looks fine, but it just really irked me that there was no way for customers to write something effective using our insights library and we can do better.

PR #491, opened 2023-08-30
chore(deps): unpin some dependencies

benhinchley:
This PR aims to unpin some of our dependencies so that we can reduce the [number of security alerts](https://github.com/algolia/search-insights.js/security/dependabot) and also eventually get up to latest for all dependencies (which would allow us to enable dependabot).

Notes:
- Upgrading Typescript above 4.x breaks tslint
- Upgrading Prettier above 2.x breaks tslint
- Upgrading Jest seems to break the tests and build


jonathaningram (changes requested):
Looks like the build failed?

benhinchley:
> Looks like the build failed?

Oh yeah I'm having a tonne of fun with that at the moment, build works fine locally..........

jonathaningram (approved):

PR #494, opened 2023-08-31
chore(deps): upgrade to latest versions of all deps

benhinchley:
This PR migrates us from ts-lint to eslint (using https://github.com/algolia/eslint-config-algolia), and updates the rest of the dependencies to their latest versions.


.prettierrc
```diff
@@ -1,5 +1,4 @@
 {
-  "singleQuote": false,
-  "semi": true,
-  "trailingComma": "none"
-}
+  "singleQuote": true,
```
jonathaningram, line 2: I think it would be helpful to leave as double quote for this PR so it's easier to read the diff if that's OK?

benhinchley: eslint-config-algolia enforces singleQuotes.... I guess we can disable that rule.

JasonBerry: Do it in a subsequent PR maybe?

benhinchley: Yeah all good, figured out how to make eslint honor the local prettier config

lib/polyfills/objectAssign.ts
JasonBerry, line 1: This whole polyfill can go. `Object.assign` has been available for a long time now.

lib/polyfills/objectKeys.ts
JasonBerry, line 1: Kill this polyfill too.

benhinchley:
@JasonBerry or @jonathaningram if either of you have any time to look into the test failures, would be greatly appreciated. I've got to switch to something else for the afternoon.

package.json
```diff
@@ -22,10 +22,10 @@
   "scripts": {
     "build": "rollup --environment NODE_ENV:'production' -c rollup.config.mjs && tsc --build tsconfig.declaration.json",
     "build:dev": "rollup --watch --environment NODE_ENV:'development' -c rollup.config.mjs",
-    "lint": "tslint ./lib/**/*.ts",
-    "lint:fix": "yarn lint -- --fix",
+    "lint": "eslint --ext .js,.ts,.tsx lib",
+    "lint:fix": "npm run lint -- --fix",
     "prettier": "prettier 'lib/**/*.ts' 'docs/**/*.md' --write",
-    "test": "jest",
+    "test": "jest --runInBand",
```
JasonBerry, line 28: Oddly this change resulted in the tests finishing _faster_. 🤷‍♂️

JasonBerry (approved):

package.json
jonathaningram: Were you able to discover the underlying issue? Is it just that the tests are designed in such a way that they can't be done in parallel?

jonathaningram (approved):

package.json
JasonBerry: I tried a few different things but could not figure it out. I thought that workers might have been getting held open due to our use of XHR (and so jsdom's implementation of it), but waiting on those requests in the tests didn't fix anything and made them take a lot longer.

benhinchley:
🚀

PR #508, opened 2023-09-05
fix: add missing type defs for conversion subtypes

benhinchley:
The new conversion subtypes added a top level value and currency attribute to the event object, as well as allowing a queryID to be provided via the objectData.

EEX-703

### DIFF
--- a/lib/types.ts
+++ b/lib/types.ts
@@ -85,6 +85,8 @@ export type InsightsEventType = "click" | "conversion" | "view";
 export type InsightsEventConversionSubType = "addToCart" | "purchase";
 
 export type InsightsEventObjectData = {
+  queryID?: string;
+
   price?: number;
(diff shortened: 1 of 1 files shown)


lib/types.ts
```diff
@@ -85,6 +85,8 @@ export type InsightsEventType = "click" | "conversion" | "view";
 export type InsightsEventConversionSubType = "addToCart" | "purchase";
 
 export type InsightsEventObjectData = {
+  queryID?: string;
+
   price?: number;
```
soulplant, line 90: price should also accept a string (and so should discount) in order for callers to be able to provide us with precise decimal numbers, rather than floats.

Same applies to top level `value`.

soulplant (changes requested):

soulplant (approved):

PR #513, opened 2023-09-27
feat: add support for authenticatedUserToken

jkaho:
### What
* Add `authenticatedUserToken` field to event types
* Add `setAuthenticatedUserToken` method to enable users to set the field to be automatically passed in subsequent `sendEvent` calls
* Add `getAuthenticatedUserToken` method to enable users to get the `authenticatedUserToken` field
* Add `onAuthenticatedUserTokenChange` to enable passing a callback on `authenticatedUserToken` change
* Enable `authenticatedUserToken` to be passed on `init`

### Links
[EEX-746](https://algolia.atlassian.net/browse/EEX-746)

More context:
* [Insights user identity proposal (Confluence)](https://algolia.atlassian.net/wiki/spaces/PREDICT/pages/4559470786/Insights+user+identity+proposal)
* [Authenticated user token in Insights pipeline (Confluence)]( https://algolia.atlassian.net/wiki/spaces/EX/pages/4576346340/2023-05-25+Authenticated+user+token+in+Insights+pipeline)

[EEX-746]: https://algolia.atlassian.net/browse/EEX-746?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -29,7 +29,7 @@ aa("init", {
 
 ## 3. Add `userToken`
 
-On the Node.js environment, unlike the browser environment, a [user token](https://www.algolia.com/doc/guides/sending-events/concepts/usertoken) (`userToken` and/or `authenticatedUserToken`) must be specified when sending any event.
+On the Node.js environment, unlike the browser environment, a [user token](https://www.algolia.com/doc/guides/sending-events/concepts/usertoken) (required `userToken` and optional `authenticatedUserToken`) must be specified when sending any event.
 
 ```js
 // Anonymous user ID
@@ -40,6 +40,7 @@ aa("clickedObjectIDs", {
 
 // Authenticated user ID
 aa("clickedObjectIDs", {
+  userToken: "ANONYMOUS_ID",
   authenticatedUserToken: "USER_ID"
   // ...
 });
--- a/docs/nodejs.md
+++ b/docs/nodejs.md
@@ -29,11 +29,18 @@ aa("init", {
 
 ## 3. Add `userToken`
 
-On the Node.js environment, unlike the browser environment, `userToken` must be specified when sending any event.
+On the Node.js environment, unlike the browser environment, a [user token](https://www.algolia.com/doc/guides/sending-events/concepts/usertoken) (`userToken` and/or `authenticatedUserToken`) must be specified when sending any event.
 
 ```js
+// Anonymous user ID
 aa("clickedObjectIDs", {
-  userToken: "USER_ID"
+  userToken: "ANONYMOUS_ID"
+  // ...
+});
+
+// Authenticated user ID
+aa("clickedObjectIDs", {
+  authenticatedUserToken: "USER_ID"
   // ...
 });
 ```
--- a/lib/__tests__/_sendEvent.test.ts
+++ b/lib/__tests__/_sendEvent.test.ts
@@ -403,7 +403,7 @@ describe("sendEvents", () => {
       });
     });
 
-    it("should be added by default", () => {
+    it("should be added by default if authenticatedUserToken not provided", () => {
       expect(analyticsInstance._anonymousUserToken).toBe(true);
 
       analyticsInstance.sendEvents(
@@ -490,6 +490,41 @@ describe("sendEvents", () => {
           expect.objectContaining({
             userToken: "my-user-token"
           })
+        ]
+      });
+    });
+
+    it("should not be added if authenticatedUserToken is provided", () => {
+      analyticsInstance.setAuthenticatedUserToken("my-user-token");
+
+      analyticsInstance.sendEvents(
+        [
+          {
+            eventType: "click",
+            eventName: "my-event",
+            index: "my-index",
+            objectIDs: ["1"]
+          }
+        ],
+        {
+          headers: {
+            "X-Algolia-Application-Id": "algoliaAppId",
+            "X-Algolia-API-Key": "algoliaApiKey"
+          }
+        }
+      );
+      expect(XMLHttpRequest.send).toHaveBeenCalledTimes(1);
+      const payload = JSON.parse(XMLHttpRequest.send.mock.calls[0][0]);
+      expect(payload).toEqual({
+        events: [
+          {
+            eventType: "click",
+            eventName: "my-event",
+            index: "my-index",
+            objectIDs: ["1"],
+            authenticatedUserToken: "my-user-token",
+            userToken: undefined
+          }
         ]
       });
     });
--- a/lib/__tests__/_sendEvent.test.ts
+++ b/lib/__tests__/_sendEvent.test.ts
@@ -495,6 +495,82 @@ describe("sendEvents", () => {
     });
   });
 
+  describe("authenticatedUserToken", () => {
+    let analyticsInstance: AlgoliaAnalytics;
+    beforeEach(() => {
+      analyticsInstance = setupInstance();
+    });
+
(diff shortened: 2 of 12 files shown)


lib/_tokenUtils.ts
```diff
@@ -68,6 +68,17 @@ export function setUserToken(
...
+  this: AlgoliaAnalytics,
+  authenticatedUserToken: number | string
+): number | string {
+  this._authenticatedUserToken = authenticatedUserToken;
+  if (isFunction(this._onAuthenticatedUserTokenChangeCallback)) {
+    this._onAuthenticatedUserTokenChangeCallback(this._authenticatedUserToken);
+  }
+  return this._authenticatedUserToken;
+}
+
```
JasonBerry, line 81: I think that we should add a getter too, as we already have it for `userToken`

JasonBerry (commented):
In a similar manner to [Segment's](https://segment.com/docs/connections/spec/best-practices-identify/) `identify` method I feel like we need a way to easily `reset` the identifiers. I know we're not at the point of adding an "identify" type event at the moment, but I just wanted to raise it.

lib/_tokenUtils.ts
jkaho: My thinking for leaving it out was that the `authenticatedUserToken` is always provided, unlike the `userToken` that may be set by us, so I wasn't sure that a getter was necessary. Though, I can see how it might be useful if they were to lose the token on their end somehow. Or maybe even just to check that it's been set properly?

JasonBerry: I don't know exactly how or why it would be used, but I'd bet we'd have a feature request for it in fairly short order if it wasn't available.

JasonBerry (approved):

JasonBerry (approved):

francoischalifour:
> In a similar manner to [Segment's](https://segment.com/docs/connections/spec/best-practices-identify/) `identify` method I feel like we need a way to easily `reset` the identifiers. I know we're not at the point of adding an "identify" type event at the moment, but I just wanted to raise it.

Definitely! I wrote a proposal about accepting `aa('setUserToken', undefined)` (or `aa('setAuthenticatedUserToken', undefined)`) to restore the user token in [**RFC: Ability to restore anonymous user token in Search Insights**](https://algolia.atlassian.net/wiki/spaces/PREDICT/pages/4594499885/RFC+Ability+to+restore+anonymous+user+token+in+Search+Insights).

Although it doesn't have to be implemented at the same time as this introduction, it's needed to support logout scenarios.

jkaho:
@JasonBerry made some additional changes as per my commits 🙏

docs/nodejs.md
```diff
@@ -27,17 +27,6 @@ aa("init", {
 });
 ```
 
-## 3. Add `userToken`
```
jkaho, line 30: @JasonBerry do you know if this holds true? I created a tiny Node.js app that simply sends an Insights event and setting `userToken` on the event object didn't seem necessary. Let me know if you can think of a reason why we might require this

JasonBerry: Did the event validate once it hit Algolia or did it fail? I would expect it to fail. I believe we _want_ to remove the necessity of a user token being present on the backend for clicks but I don't think that work has been completed.

jkaho: As per Jason's Slack DM, there could be a good reason for this, e.g., anon ID might be generated per request in the Node env, so have added the text back in for now to err on the safe side. Also updated it to mention `authenticatedUserToken`.

lib/__tests__/_tokenUtils.test.ts
```diff
@@ -104,4 +120,28 @@ describe("tokenUtils", () => {
...
+      expect(authenticatedUserToken).toBeUndefined();
+    });
+    it("should return current authenticatedUserToken", () => {
+      analyticsInstance.setAuthenticatedUserToken("008");
+      const authenticatedUserToken =
+        analyticsInstance.getAuthenticatedUserToken();
+      expect(authenticatedUserToken).toEqual("008");
+    });
+    it("should accept a callback", () => {
+      analyticsInstance.setAuthenticatedUserToken("009");
```
JasonBerry, line 137: If the callback is never called this test will still pass. You need to add something like `expect.assertions(2);` to ensure that the expected assertions are made.

lib/__tests__/_tokenUtils.test.ts
```diff
@@ -104,4 +120,28 @@ describe("tokenUtils", () => {
...
+      analyticsInstance.getAuthenticatedUserToken(
+        {},
+        (err, authenticatedUserToken) => {
+          expect(err).toEqual(null);
+          expect(authenticatedUserToken).toEqual("009");
+        }
+      );
+    });
```
JasonBerry, line 145: If the callback is never called this test will still pass. You need to add something like `expect.assertions(2);` to ensure that the expected assertions are made.

lib/__tests__/init.test.ts
```diff
@@ -497,4 +519,60 @@ describe("init", () => {
...
+    it("should not set authenticatedUserToken if not passed", () => {
+      analyticsInstance.init({
+        apiKey: "***",
+        appId: "XXX"
+      });
+      expect(setAuthenticatedUserToken).toHaveBeenCalledTimes(0);
+      analyticsInstance.getAuthenticatedUserToken(null, (_err, value) => {
+        expect(value).toBeUndefined();
+      });
+    });
```
JasonBerry, line 576: You'll need to use `expect.assertions(...);` on these too

JasonBerry (changes requested):
A couple of changes required on the tests.

lib/__tests__/_tokenUtils.test.ts
jkaho: Ah TIL 🚀 Thanks!

JasonBerry (approved):

docs/nodejs.md
```diff
@@ -29,11 +29,18 @@ aa("init", {
...
+// Anonymous user ID
 aa("clickedObjectIDs", {
-  userToken: "USER_ID"
+  userToken: "ANONYMOUS_ID"
+  // ...
+});
+
+// Authenticated user ID
+aa("clickedObjectIDs", {
+  authenticatedUserToken: "USER_ID"
```
soulplant, line 43: I think we still want users to provide `userToken` here in addition to `authenticatedUserToken`.

lib/__tests__/_sendEvent.test.ts
```diff
@@ -493,6 +493,117 @@ describe("sendEvents", () => {
         ]
       });
     });
+
+    it("should not be added if authenticatedUserToken is provided", () => {
```
soulplant, line 497: Does this need to change in light of the discussion here? https://algolia.atlassian.net/wiki/spaces/EX/pages/4576346340/2023-05-25+Authenticated+user+token+in+Insights+pipeline?focusedCommentId=4746707079

soulplant (changes requested):

lib/__tests__/_sendEvent.test.ts
jkaho: Yes, good timing, just pushed the change!

lib/__tests__/_sendEvent.test.ts
jkaho, line 1: Yikes sorry rearranged the tests a bit which has made it a bit annoying to read

soulplant (approved):

Issue #444, opened 2023-04-19
Support Edge runtime alongside Node and Browser


MaxLeiter:
The [getRequestForNode](https://github.com/algolia/search-insights.js/blob/886fc6191bef4fafeb8097f8ac5f37164137c5bc/lib/utils/getRequesterForBrowser.ts#L13)/Browser utilities fail when used in an [edge runtime](edge-runtime.vercel.app).

The following is logged when it's attempted: 
```
[Error: Could not find a supported HTTP request client in this environment.]
```

Jeffrey-Zutt:
We are running into the exact same issue.

benhinchley:
fixed by #515

PR #515, opened 2023-10-30
feat: introduce native fetch fallback when http-module is not supported

Jeffrey-Zutt:
First of all, thank you for maintaining the great library!

As described in issue #444, importing `search-insights` on an '[edge-runtime](https://nextjs.org/docs/app/building-your-application/rendering/edge-and-nodejs-runtimes)' results in the following error: 
`[Error: Could not find a supported HTTP request client in this environment.]`

The edge runtime however, does [support `fetch`](https://nextjs.org/docs/pages/api-reference/edge) (and is added to node [per version 18](https://nodejs.org/dist/latest-v18.x/docs/api/globals.html#fetch)).

This PR adds a fallback to the native `fetch` method when nodeHttpRequest is not available.

Please let me know if you'd like to see anything changed :-).

Kind regards.

### DIFF
--- a/jest.setup.js
+++ b/jest.setup.js
@@ -1,1 +1,3 @@
 console.info = () => {};
+
+require("jest-fetch-mock").enableMocks();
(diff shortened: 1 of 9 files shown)


benhinchley (approved):
Thanks for taking the time to make this change 😄 🚀

PR #522, opened 2023-11-27
chore: deprecate purchase event top-level queryID param

jkaho:
Create distinct type for `purchase` event parameters - same as a normal `conversion` event but the top-level `queryID` is optional and deprecated.

[EEX-808](https://algolia.atlassian.net/browse/EEX-808)

[EEX-808]: https://algolia.atlassian.net/browse/EEX-808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/lib/conversion.ts
+++ b/lib/conversion.ts
@@ -27,25 +27,12 @@ export function convertedObjectIDsAfterSearch(
   return this.sendEvents(addEventType("conversion", events), additionalParams);
 }
 
-export interface InsightsSearchConversionSubtypeEvent {
-  eventName: string;
-  userToken?: string;
-  authenticatedUserToken?: string;
-  timestamp?: number;
-  index: string;
-
-  queryID?: string;
-  objectIDs: string[];
-  objectData?: InsightsEvent["objectData"];
-  currency?: InsightsEvent["currency"];
-}
-
 export function addedToCartObjectIDsAfterSearch(
   this: AlgoliaAnalytics,
-  ...params: Array<WithAdditionalParams<InsightsSearchConversionSubtypeEvent>>
+  ...params: Array<WithAdditionalParams<InsightsSearchConversionEvent>>
 ): ReturnType<AlgoliaAnalytics["sendEvents"]> {
   const { events, additionalParams } =
-    extractAdditionalParams<InsightsSearchConversionSubtypeEvent>(params);
+    extractAdditionalParams<InsightsSearchConversionEvent>(params);
 
   return this.sendEvents(
     addEventTypeAndSubtype("conversion", "addToCart", events),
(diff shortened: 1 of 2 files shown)


jkaho:
@soulplant and/or @benhinchley not sure that this is necessary or the right fix, just seemed like the lowest impact one to avoid confusion around customers needing to set the top-level `queryID` for `addToCart` and `purchase` events.

This would probably make it more likely for people to send the events without a query ID which defeats the purpose of `purchasedObjectIDsAfterSearch` and `addedToObjectIDsAfterSearch`, but I wasn't sure that we'd want to go down the path of making `objectData` and `objectData.queryID` required for these two methods. Thoughts?

soulplant:
Good thought, Jess.

As per [the spec](https://algolia.atlassian.net/wiki/spaces/EX/pages/4652531940/Insights+update#Purchase.1) I'm not sure if we want purchase and addToCart events to be the same in this regard. addToCart allows a top-level queryID, but purchase doesn't.

In both cases the queryID (be it the top level queryID or objectData.queryID) are optional.

So I guess the change we would want to make is to remove the top level queryID for purchase events only. I think the rationale here is that addToCart is often done in the context of the 'current query' (e.g. adding to cart directly from the search results page), whereas purchase is usually done after there is no current query and will vary for each item. @jkaho you might want to confirm with @benfranz if that's the right interpretation (not sure if he'll see GH mentions).

jkaho:
@soulplant ah okay I think I may have misunderstood. Didn't realise `addToCart` and `purchase` were different, thanks for clarifying! I've pinged Ben, we'll see what he says.

jkaho:
> So I guess the change we would want to make is to remove the top level queryID for purchase events only. I think the rationale here is that addToCart is often done in the context of the 'current query' (e.g. adding to cart directly from the search results page), whereas purchase is usually done after there is no current query and will vary for each item.

Ben's response:
> Yes, this is only for purchase events.
>
> Add to cart events have an event-level queryID, just like generic conversions.
> 
> That’s because there can always only be one queryID per addToCart event.
> 
> But a purchase event has n queryIDs.
> 
> You can add sth to your cart. Then search another query and add sth else to your cart… then, when you hit the purchase button, you have two different objects in your cart that came from two different queryIDs.

jkaho:
@soulplant would this be considered a breaking change?

JasonBerry:
> @soulplant would this be considered a breaking change?

IMO, yes. I'm pretty sure that if customers update to this version they'll get TS errors if they're specify a `queryID` at the top level.

jkaho:
@JasonBerry yeah, I was thinking the same, which is why I initially thought the lower impact change here would be to make the param optional rather than removing it entirely. We could keep the param in and not mention it in the docs; simply instruct users to set the object-level query IDs. Do you have any thoughts or preferences on how we should proceed?

JasonBerry:
> Do you have any thoughts or preferences on how we should proceed?

Make it optional, mark it as `@deprecated`, apply any `queryID` that was supplied at the top level to the items?

jkaho:
> apply any `queryID` that was supplied at the top level to the items in objectData

@JasonBerry my understanding is that we already do this on the backend if the object level `queryID` isn't provided, i.e., we haven't actually deprecated the field, we're just wanting to change the way we present the schema to customers to avoid confusion about how to set the query ID.

@soulplant can you confirm whether this is necessary?

soulplant:
@jkaho Yep, it shouldn't be necessary to set the top level queryID on objectData[].queryID - as you say the backend already interprets it this way.

lib/conversion.ts
```diff
@@ -40,12 +40,20 @@ export function addedToCartObjectIDsAfterSearch(
   );
 }
 
+export type InsightsSearchPurchaseEvent = Omit<
+  InsightsSearchConversionEvent,
+  "queryID"
+> & {
+  /** @deprecated Use objectData.queryID instead. */
```
soulplant, line 47: Nice!

soulplant (approved):

PR #534, opened 2024-06-25
feat: add query id inference for conversion events based on prior clicks

JasonBerry:
This is the majority of the inference logic. A couple of things not added yet:
* when `eventSubtype` is `addToCart` store any queryID passed for future `purchase` conversion events
* removal of objectID/queryID association from store when `eventSubtype` is `purchase`
* pruning of associations older than 90 days

### DIFF
--- a/lib/__tests__/_sendEvent.test.ts
+++ b/lib/__tests__/_sendEvent.test.ts
@@ -1,12 +1,18 @@
-import * as querystring from "querystring";
-import * as url from "url";
-
 import AlgoliaAnalytics from "../insights";
+import { storeQueryForObject } from "../utils";
 import { getRequesterForBrowser } from "../utils/getRequesterForBrowser";
 
 jest.mock("../../package.json", () => ({
   version: "1.0.1"
 }));
+
+function searchParamsToObject(params: URLSearchParams) {
+  const result: { [key: string]: string } = {};
+  for (const [key, value] of params.entries()) {
+    result[key] = value;
+  }
+  return result;
+}
 
 const credentials = {
   apiKey: "testKey",
@@ -27,6 +33,7 @@ describe("sendEvents", () => {
   let XMLHttpRequest: { open: any; send: any };
 
   beforeEach(() => {
+    localStorage.clear();
     XMLHttpRequest = {
       open: jest.spyOn((window as any).XMLHttpRequest.prototype, "open"),
       send: jest.spyOn((window as any).XMLHttpRequest.prototype, "send")
@@ -62,7 +69,7 @@ describe("sendEvents", () => {
       expect(XMLHttpRequest.open).toHaveBeenCalledTimes(1);
       const [verb, requestUrl] = XMLHttpRequest.open.mock.calls[0];
       expect(verb).toBe("POST");
-      expect(url.parse(requestUrl).pathname).toBe("/1/events");
+      expect(new URL(requestUrl).pathname).toBe("/1/events");
     });
     it("should pass over the payload with multiple events", () => {
       analyticsInstance.sendEvents([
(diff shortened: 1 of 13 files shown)


lib/click.ts
```diff
@@ -22,6 +22,12 @@ export function clickedObjectIDsAfterSearch(
   const { events, additionalParams } =
     extractAdditionalParams<InsightsSearchClickEvent>(params);
 
+  events.forEach(({ index, queryID, objectIDs }) =>
```
jkaho, line 25: Just checking that we want to store query IDs even if the user hasn't opted in the query ID inference? I guess this might make it more effective if they decide to enable it?

lib/utils/objectQueryTracker.ts
```diff
@@ -0,0 +1,32 @@
...
+}
+
+export function storeQueryForObject(
+  index: string,
+  objectID: string,
+  queryID: string
+): void {
+  const objectQueryMap = getCache(index);
+
+  objectQueryMap[objectID] = [queryID, Date.now()];
```
jkaho, line 23: Sorry if obvious, but what's the reason we're not using the event timestamp?

Is it worth checking for existing stored query IDs and comparing timestamps before setting a new queryID-objectID association? E.g., is it possible that a customer might pass a `timestamp` for a click event that came _before_ the timestamp stored in local storage? I assume we'd want the latest one.

lib/click.ts
JasonBerry: There's no downside to storing it, and as you say once it's enabled it would instantly begin to work.

lib/utils/objectQueryTracker.ts
JasonBerry: It's theoretically _possible_ that someone might fire a click on the frontend with an older `timestamp` but I think that would be pretty unlikely. Anyone passing a timestamp to a search-insights event is most likely going to be using `Date.now()` themselves.

Really I'm just adding a timestamp here so that I know when to nuke it. The last `queryID` to be set always overwrites the existing one.

Or am I missing a pretty common usecase?

lib/__tests__/_sendEvent.test.ts
```diff
@@ -93,13 +100,62 @@ describe("sendEvents", () => {
...
-      const { query } = url.parse(requestUrl);
-      expect(querystring.parse(query!)).toEqual({
+      const { searchParams } = new URL(requestUrl);
+      expect(searchParamsToObject(searchParams)).toEqual({
         "X-Algolia-API-Key": "testKey",
         "X-Algolia-Agent": "insights-js (1.0.1); insights-js-node-cjs (1.0.1)",
         "X-Algolia-Application-Id": "testId"
       });
     });
+    it("should infer query ids when additionalParams.inferQueryID is true", () => {
```
jkaho, line 110: Should we add the inverse? I.e., if `inferQueryID` is false?

JasonBerry: Yep, I'll add that in a follow-up PR.

lib/utils/localStorage.ts
```diff
@@ -0,0 +1,66 @@
...
+  static remove(key: string): void {
+    localStorage.removeItem(key);
+  }
+
+  /**
+   * Check if the storage is nearly full.
+   *
+   * @returns A promise that resolves to a boolean indicating if the storage is > 90% full.
+   */
+  static isNearlyFull(): Promise<boolean> {
```
jkaho, line 59: Quick question - I presume it'll be done in a subsequent change but when do we intend to call this?

JasonBerry: Following PR will have this.

jkaho (approved):

PR #535, opened 2024-07-04
feat: query id inference pruning of purchased and oldest associations when limit is reached

JasonBerry:
Added the remaining logic to:
* store any queryID passed for future `purchase` conversion events when `eventSubtype` is `addToCart`
* remove objectID/queryID association from store when `eventSubtype` is `purchase`
* keep most recent 4,000 associations when store reaches 5,000

### DIFF
--- a/lib/_sendEvent.ts
+++ b/lib/_sendEvent.ts
@@ -1,7 +1,7 @@
 import { addQueryId } from "./_addQueryId";
 import type AlgoliaAnalytics from "./insights";
 import type { InsightsAdditionalEventParams, InsightsEvent } from "./types";
-import { isUndefined } from "./utils";
+import { isPromise, isUndefined, removeQueryForObjects } from "./utils";
 import type { RequestFnType } from "./utils/request";
 
 export function makeSendEvents(requestFn: RequestFnType) {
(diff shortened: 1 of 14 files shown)


lib/_sendEvent.ts
```diff
@@ -57,6 +57,25 @@ export function makeSendEvents(requestFn: RequestFnType) {
       this._apiKey,
       additionalParams?.headers
     );
+    return isPromise(send) ? send.then(purgePurchased(events)) : send;
```
jkaho, line 60: Just so I'm clear, what's the reasoning behind not purging in `purchasedObjectIDs(AfterSearch)` seeing as, if I'm understanding correctly, that's the level of abstraction we're doing the storing?

JasonBerry: I made the choice to put it here because once an object is purchased (no matter the method used to mark the object as purchased) I want it removed from the object/query cache. Our customers may use either `purchasedObjectIDs` or `purchasedObjectIDsAfterSearch` (they could even use `convertedObjectIDs` or `convertedObjectIDsAfterSearch` if they're not using TS and it would still work if they're passing in the `eventSubtype` param) and I wanted to ensure it covered all of these cases.

jkaho (approved):

PR #562, opened 2024-10-28
fix: save existing anon tokens for multiple init calls 

shaejaz:
### Context

In InstantSearch, in order for user tokens to be set in time for the first query on page load, anonymous user tokens are generated and set as both the InstantSearch token and Insights token.

This flow has one issue though. When `useCookies` flag needs to be changed later (such as with a cookie consent banner), a second `init` call to insights with the flag set to true does not save the existing token to a cookie.

A short repro is as follows with the very latest version of InstantSearch where this behaviour was added (PR: https://github.com/algolia/instantsearch/pull/6377):

```js
const searchClient = algoliasearch(
  // ...
);

window.aa('init', {
  appId: 'id',
  apiKey: 'key',
});

const search = instantsearch({
  indexName: 'index',
  searchClient,
  insights: true,
});

search.addWidgets([
  // ...
]);

// This does not save the cookie atm
setTimeout(() => {
  window.aa('init', {
    partial: true,
    useCookie: true,
  });
}, 5000);
```

### DIFF
--- a/lib/__tests__/init.test.ts
+++ b/lib/__tests__/init.test.ts
@@ -240,6 +240,23 @@ describe("init", () => {
 
     setUserToken.mockRestore();
   });
+  it("should save anonymous userToken as cookie when useCookie is set to true later", () => {
+    analyticsInstance.init({
+      apiKey: "***",
+      appId: "XXX"
+    });
+
+    analyticsInstance.setUserToken("anonymous-123123123");
+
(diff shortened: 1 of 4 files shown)


Haroenv (approved):
from the InstantSearch point of view, this is good

JasonBerry (approved):

PR #567, opened 2025-03-11
chore(deps): upgrade rollup and plugins

JasonBerry:
Switches to terser instead of uglify for minification.


jkaho (commented):
Out of interest, what prompted this change?

JasonBerry:
> Out of interest, what prompted this change?

Was a branch I had lying around for a while to modernise some of the dependencies, a number of which have security warnings (some are many years old and no longer receiving updates). Migrating to rollup 4 (released 18 months ago) takes [a few more steps](https://rollupjs.org/migration/), so this is just an interim to move us closer.

jkaho (approved):

